Issue #169, opened 2020-04-09
Add Prettier as mandatory check in CI Pipeline


palexster:
Prettier ensures that the TS code is formatted following the project specifications. It should be added as mandatory check when proposing a new PR.
Items:

- [x] Add the check to a new Github check ```./node_modules/.bin/prettier --check */**/*.{js,json,css}```
- [x] Document how to run the prettify over the project ```npm run prettify```
- [x] Document how to enable git commit pre-hook

palexster:
Closing.

PR #198, opened 2020-04-16
Added lint check action in webservice

GabriFila:
This PR contains the resolution for #169 :
- added package.json script to check linting
- created YAML file with action

**_Note_**: in the YAML file at line 15-16 I decided to install prettier only, this brings advantages by reducing the time needed by the action. But it requires that, when we need other packages for linting, we'll have to add them also in the action.


palexster:
Very nice!

PR #204, opened 2020-04-18
Add Configmap to operator deployment

palexster:
Add configmap to kustomization.yaml to complete the labOperator deployment

### DIFF
--- a/operators/labInstance-operator/k8s/operator/configmap.yaml
+++ b/operators/labInstance-operator/k8s/operator/configmap.yaml
@@ -1,0 +1,11 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: operator-config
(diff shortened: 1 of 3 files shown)


operators/labInstance-operator/k8s/operator/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: operator-config
+  namespace: lab-operator
+data:
+  WebdavSecretName: nextcloud-credentials
```
giorio94, line 7: I would change WebdavSecretName to webdavSecretName (first letter lower case), in order to uniform this name with the others and avoid future problems. The same modification should be performed in the deployment.

PR #206, opened 2020-04-18
Pipeline Refactoring and Manifest Improvement

palexster:
- [x] Separate  Pipelines for Laboratory Operator and Frontend
- [x] Kube-score step to assess quality of K8s manifests
- [x] Adding readiness and liveness check to laboperator and frontend containers
- [x] Add security context to k8s manifests
- [x] Add requests and limits to k8s manifests
- [x] The user used in container is no more root, but a normal user\
- [x] Add stable tag to docker images build in release pipeline

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -1,0 +1,52 @@
+name: Frontend 
+on: [push]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Website - KubeScore Check
+      run: |
+           docker run -v  ${{ github.workspace }}:/Crownlabs zegl/kube-score score --ignore-test=pod-networkpolicy,container-security-context /Crownlabs/webservice/kubernetes/deployment.yaml /Crownlabs/webservice/kubernetes/svc.yaml /Crownlabs/webservice/kubernetes/ingress-callback.yaml  /Crownlabs/webservice/kubernetes/ingress-content.yaml
+
+    - name: Build and Push Frontend image (Dev)
+      uses: elgohr/Publish-Docker-Github-Action@master
+      with:
+        name: crownlabs/frontend
+        username: ${{ secrets.DOCKER_USERNAME }}
+        dockerfile: Dockerfile
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        workdir: webservice
+        tags: "latest,${{ github.sha }}"
+      if: github.ref != 'refs/heads/master'
+
+    - name: Build and Push Frontend image
+      uses: elgohr/Publish-Docker-Github-Action@master
+      with:
+        name: crownlabs/frontend
+        username: ${{ secrets.DOCKER_USERNAME }}
+        dockerfile: Dockerfile
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        workdir: webservice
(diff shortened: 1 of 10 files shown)


.github/workflows/frontend.yml
```diff
@@ -7,25 +7,22 @@ jobs:
...
-        fetch-depth: 0
-        token: ${{ secrets.CI_TOKEN }}
-    - name: Build Laboratory-Operator image
-      run: cd operators/labInstance-operator/ && docker build -f Dockerfile . 
-      if: github.ref != 'refs/heads/master'
-    - name: Build and Push Laboratory-Operator image
+
+    - name: Website - KubeScore Check
+      run: |
+           docker run -v  ${{ github.workspace }}:/Crownlabs zegl/kube-score score --ignore-test=pod-networkpolicy,container-security-context /Crownlabs/webservice/kubernetes/deployment.yaml /Crownlabs/webservice/kubernetes/svc.yaml /Crownlabs/webservice/kubernetes/ingress-callback.yaml  /Crownlabs/webservice/kubernetes/ingress-content.yaml
```
giorio94, line 13: Would it be possible to specify a wildcard such as `/Crownlabs/webservice/kubernetes/*.yaml`? Otherwise I fill we will never update this upon the addition of new manifests.

.github/workflows/laboratory-operator.yml
```diff
@@ -0,0 +1,62 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: LabOperator - KubeScore Check
+      run: |
+       docker run -v  ${{ github.workspace }}/operators/labInstance-operator/:/labInstance-operator zegl/kube-score score --ignore-test=pod-networkpolicy /labInstance-operator/k8s/operator/deployment.yaml
+
+    - name: Build and Push Laboratory-Operator image
```
giorio94, line 32: I would add (Dev) to the name for the sake of clarity (as in the frontend file).

.github/workflows/frontend.yml
```diff
@@ -7,25 +7,22 @@ jobs:
...
+
+    - name: Build and Push Frontend image (Dev)
```
mLavacca, line 15: Shouldn't this image be the pre-prod version instead of dev?

palexster: AFAIK, the image used in pre-prod are images built from master branch. This image is used only by developers pipelines.

.github/workflows/frontend.yml
palexster: Yeah, I tried with *.yaml in a previous build but it didn't work.

.github/workflows/frontend.yml
mLavacca: Yes, you are right, I didn't notice the if condition ``github.ref != 'refs/heads/master'``. Maybe in the future, we could move this step to the PR pipeline,  for the sake of clarity, if you agree.

.github/workflows/frontend.yml
palexster: I try it again and see if it was another problem.

palexster: https://github.com/netgroup-polito/CrownLabs/pull/206/checks

mLavacca: maybe the ``docker run`` can be used in a bash script that iterates over the ``/Crownlabs/webservice/kubernetes`` directory and groups all the file names in a variable

PR #221, opened 2020-04-22
CRDtemplates Management

michele6000:
- Add button to create template and destroy


michele6000:
- Fixed index on template select on professor view, now professor can manage template only if he has admin role in template's namespace

giorio94:
@michele6000 In order to simplify the review process of this pretty big PR, please:
1. Squash all commits into just one, unless relevant (this will also remove the merge commits that should not be there). Ensure that the resulting commit message is meaningful.
2. Remove any useless code (e.g commented out or `console.log`) introduced by the PR. It adds no real value but drowns the real changes.
3. Ensure that all new code is correctly formatted; at a first glance, I spotted some changes due to different prettifier settings (not actually sure which version is correct).

PR #222, opened 2020-04-22
Get branch author

mlavacca:
# Description

Before this PR, the owner of each PR was considered as the entity that applied the last modification.
Now the owner of the branch is the author of the first commit of the branch related to the PR. This is needed by the dev event dispatcher, which triggers a new deployment creation through CrownOps in a namespace related to the PR owner.

### DIFF
--- a/.github/workflows/dev-deployment.yml.disabled
+++ b/.github/workflows/dev-deployment.yml.disabled
@@ -13,11 +13,27 @@ jobs:
     name: "dev deployment event"
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@v2
+
+    - name: "make scripts executable"
+      run: chmod +x ./ci-scripts/*
+
+    - name: "set PR number"
+      run: echo "::set-env name=PR_number::`echo $GITHUB_REF | cut -f 3 -d /`"
+
+    - name: "install python dependencies"
+      run: pip install requests
+
+    - name: "set branch owner"
+      run: echo "::set-env name=branch_owner::`./ci-scripts/get_branch_owner.py $PR_number`"
+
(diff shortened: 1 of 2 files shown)


palexster (approved):
/approve

giorio94 (commented):
Thanks @mLavacca for this. I believe the feature this PR is part of is very important.

Nonetheless I honestly have some reservations on how this feature is currently implemented (although I haven't figured it out completely yet and I may be wrong). My main comments are (again, not strictly related to this PR but more general):

* I would limit the generation of the frontend development environment only if the PR actually modifies frontend files, to prevent waste of resources.
* I believe this function should be coupled with the deletion counterpart (it seems to be it is implemented in part only), to avoid polluting the cluster with resources never deleted. I would avoid enabling one without the other.
* I believe it is required a better management of the digital certificates, to avoid potentially issue too many requests and get rate limited by cert-manager (proposal: a wildcard certificate *.preprod.crownlabs.polito.it synchronized between namespaces with kubed)
* I personally find very difficult to understand what a GitHub Action does in case it refers to external scripts, having to bounce between multiple files. Hence, I would avoid them as much as possible. Nonetheless, if necessary, I would keep them within the .github directory and not in the project's root.
* I don't know whether it is better to relate the resources (e.g. namespaces, urls) to the author+branch (as now) or just to the PR number. Additionally, I am not sure whether the filter by author is required or the GitHub permissions would be enough (i.e. allow if the PR is from the same repo, block in case of forks). I believe the latter case would simplify the management and reduce the possible unexpected failures.

@mLavacca @palexster What are your thoughts?

giorio94:
Closing, as superseded by the staging pipeline.

PR #228, opened 2020-04-26
setup-crownlabs-vm - optimizations

giorio94:
This PR introduces different optimizations during the configuration of the CrownLabs VMs:
- [x] Improves the NoVNC service file for better management with systemctl (e.g. stop, restart).
- [x] Configures the vncserver working directory to the user's home (i.e. the terminal starts in `~` instead of `/`)
- [x] Configures the VM to boot in text mode. Indeed, a second graphical session is started by vncserver (i.e. the one used by the final users), while the *original* one is not seen by anybody. Yet, it increases the RAM usage of an idle VM by almost 200M (350M -> 520M), as well as increases the CPU usage and the boot time.
- [x] Disables the ubuntu splash screen. As above, it is never seen by the users but consumes resources during the boot.
- [x] Removes the vbox guest additions (if present), since they are useless in KVM (apparently, it saves 20-25 seconds at boot time).

All optimizations are available when configuring a VM with the `setup-crownlabs-vm` script, while some have been backported to `prepare-vm.sh`.

### DIFF
--- a/provisioning/virtual-machines/scripts/prepare-vm.sh
+++ b/provisioning/virtual-machines/scripts/prepare-vm.sh
@@ -65,6 +65,7 @@ After=syslog.target network.target
 Type=forking
 User=${USER}
 Group=${USER}
+WorkingDirectory=${HOME}
 ExecStartPre=/bin/sh -c '/usr/bin/vncserver -kill %i > /dev/null 2>&1 || :'
 ExecStart=/usr/bin/vncserver %i -SecurityTypes None -localhost
 ExecStop=/usr/bin/vncserver -kill %i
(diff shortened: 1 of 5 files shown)


palexster:
This is huge @giorio94 ! Thanks a lot!

giorio94:
> This is huge @giorio94 ! Thanks a lot!

Thanks. The only problem is that we have to rebuild the VM images to make them effective.

I was also thinking that we may move NoVNC outside the VM to a container (it would be easier to upgrade in case of necessity). @palexster, @AndreaCossio what is your opinion? I believe that should be easy to do.

palexster:
> > This is huge @giorio94 ! Thanks a lot!
> 
> Thanks. The only problem is that we have to rebuild the VM images to make them effective.
> 
> I was also thinking that we may move NoVNC outside the VM to a container (it would be easier to upgrade in case of necessity). @palexster, @AndreaCossio what is your opinion? I believe that should be easy to do.

I was discussing with @ReddaHawk about that, in order to decouple the life cycle of the vm with NoVNC. I think it should be not too much pain also on the sig-api side (just a new side deployment).

PR #232, opened 2020-04-28
network-multitool docker image for gns3

gabrielecastellano:
- fixed polycube dependencies
- added docker image to be used within gns3 as end devices


gabrielecastellano:
> Is there any specific reason you're moving iperf/iperf3 from the first file to the second?

Polycube was failing compiling because of missing dependencies. This applied both for iperf and libpcap-dev, that were installed later for other reason. I moved these packets in the polycube dependency list so that they are installed prior compiling polycube.

PR #243, opened 2020-05-04
Add auto-labelling for automatic triage of PRs

palexster:
This PR adds a new pipeline to auto-add labels in new PRs. When a file is modified in a specific path, a corresponding label is added to contextualize the new contribution.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,22 @@
+sig-api:
+  - operators/**/*
+
+sig-ui:
+  - webservice/**/*
+
+sig-operations:
+  - infrastructure/**/*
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,18 @@
+sig-api:
+  - operators/**/*
+
+sig-ui:
+  - webservice/**/*
+
+sig-operations:
+  - infrastructure/**/*
+  - provisioning/**/*
```
giorio94, line 9: Should we also add the user-environment label to the VM provisioning part?

palexster: Right!

PR #248, opened 2020-05-10
GUI FIX

michele6000:
- Fixed text color of item of running labs list, now on light/dark mode is always black
- Add a title bar to professor view, to disambiguate professor view from student view

### DIFF
--- a/webservice/src/UserLogic.jsx
+++ b/webservice/src/UserLogic.jsx
@@ -490,9 +490,11 @@ export default class UserLogic extends React.Component {
           }
         />
         <Body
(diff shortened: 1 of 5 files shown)


GabriFila:
I like it, only thing I believe that the title in this way takes too much space, considering that probably in the future will add other things to the dashboard. I would add something to the header, another icon or change the 'user' icon based on the status: if the user is a prof show the MUI 'School' icon

michele6000:
Bug Fix:
- Lab Instance List was the same in professor view and student view. Error was in body, same function to retrive the list was passed both in admin and in student view

PR #258, opened 2020-05-18
Fix issue #257

### LINKED ISSUE #257: Selected LabInstance index
User type: Professor
Device: Any

When selecting a LabInstance sometimes, due to wrong indexes assigned to those rows, multiple laboratories are selected the same time. This happens when you are a Professor, meaning that there are both LabInstances as a normal user and as a privileged one.

ok

ko

michele6000:
When selecting a LabInstance sometimes, due to wrong indexes assigned to those rows, multiple laboratories are selected the same time. This happens when you are a Professor, meaning that there are both LabInstances as a normal user and as a privileged one.

Fixes #257


michele6000:
fixed in pr #221

PR #267, opened 2020-05-27
Improve action to lint check linting in webservice

GabriFila:
# Description

This PR improves the action to check the linting of the /webservice folder only when there are changes in the /webservice folder

# How Has This Been Tested?

It has been tested by checking the actions history
[This](https://github.com/netgroup-polito/CrownLabs/runs/776420651?check_suite_focus=true) ran
[This](https://github.com/netgroup-polito/CrownLabs/runs/776427193?check_suite_focus=true) did not run

### DIFF
--- a/.github/workflows/webservice_lint_check.yml
+++ b/.github/workflows/webservice_lint_check.yml
@@ -6,13 +6,23 @@ jobs:
     steps:
       - name: checkout code in action environment
         uses: actions/checkout@v2
+      - name: collect data about changes in path
+        uses: dorny/paths-filter@v2.0.0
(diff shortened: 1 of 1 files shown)


GabriFila:
This PR is switched to draft because at the moment Github doesn't exclude from the required checks the actions which should not run because of embedded path filters. There are some workarounds which need to view in more depth

palexster:
> 
> 
> /lgtm
> 
> @palexster Maybe we could apply the same technique also for the other pipelines to avoid rebuilding the images when not necessary. WDYT?

Totally Agree :)

PR #271, opened 2020-05-30
added Drift(Legacy) chat support extension

michele6000:
- Added possibility to support users directly via Slack
- Added analytics of users connected on website (Drift analytics)

### DIFF
--- a/webservice/src/components/Body.jsx
+++ b/webservice/src/components/Body.jsx
@@ -4,7 +4,58 @@ import IconButton from '@material-ui/core/IconButton';
 import ProfessorView from '../views/ProfessorView';
 import StudentView from '../views/StudentView';
 
+const ChatSupport = () => {
+  !(function () {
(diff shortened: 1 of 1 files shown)


webservice/src/components/Body.jsx
```diff
@@ -4,7 +4,58 @@ import IconButton from '@material-ui/core/IconButton';
...
+            (o.async = !0),
+            (o.crossorigin = 'anonymous'),
+            (o.src = `https://js.driftt.com/include/${n}/${t}.js`);
+          const i = document.getElementsByTagName('script')[0];
+          i.parentNode.insertBefore(o, i);
+        });
+    }
+  })();
+  drift.SNIPPET_VERSION = '0.3.1';
+  drift.load('47aa3t6wszpf');
```
giorio94, line 54: This appears to me to be a secret. It should not be pushed here, since the repository is public

palexster:
Closing due to the misalignment with the problem.

PR #284, opened 2020-07-11
Webservice - external kubernetes-client-javascript library

giorio94:
# Description

This PR removes the local copy of the kubernetes-client-javascript library (i.e. the webservice/k8s_library folder, hence the huge number of files deleted) in favor of the version available at [LiqoTech/kubernetes-client-javascript](https://github.com/LiqoTech/kubernetes-client-javascript), to avoid code duplication and unify the maintenance efforts.

Although the library has been configured as dependency, it is apparently not installed correctly by `npm install` and requires some manual intervention (hence the first step in the Dockerfile). @GabriFila  Could you please have a look whether this could be easily fixed? I believe it should be necessary to change a little the `package.json` file of the library but I am a totally newbie in this field. For the moment, I will keep the PR as draft.

@palexster The kube-score check started failing due to a version update changing the behavior of a test (BTW, should we specify a version tag to prevent this kind of issues in the future?). Now, according to the [kube-score suggestion](https://github.com/zegl/kube-score/blob/master/README_PROBES.md), I deleted the livenessProbe of the frontend deployment, but any alternative solution would be fine to me.

# How Has This Been Tested?

This PR does not introduce any functional changes visible to the users. It has been tested building the frontend docker image both locally (checking that the resulting image worked correctly) and with the CI (the action succeeded).


GabriFila:
The `npm install` works locally, but not in the container build. As @giorio94 found at [this link](https://npm.community/t/npm-ci-not-running-prepare-for-git-dependencies-when-run-as-root-in-docker/4485), the `unsafe-perm` must be explicitly set with the `config` command.

Also 2 more files need to be modified
- .eslintignore (update)
- ,prettierignore (delete)

giorio94:
@GabriFila Thanks! Should be fixed now. When you have time, the PR should be ready for the review (I separated the deletion commit from the other which changes the build logic for clarity).

PR #294, opened 2020-09-01
VM provisioning: add the --vbox-only flag

giorio94:
# Description

This PR merges each CrownLabs playbook with the corresponding vanilla one, introducing a `--vbox-only` flag to the `setup-crownlabs-vm` script to opt out from the installation of the CrownLabs components during the VM configuration phase (the documentation is update accordingly).
Additionally, it addresses the issues raised by the shellcheck tool.


frisso (approved):

frisso:
@giorio94 This looks very very good to me.
Just asking: can we move the requirement checking to the actual action requested to the script? This would allow the script to run on Windows in order to create the OVA. It seems to me that the only requirement that is not easy to satisfy on Windows is to have Docker running; if we move the check of Docker when it is actually needed, we can work most of the time on Windows as well. Thanks!

frisso (approved):

frisso (approved):

palexster (approved):

PR #307, opened 2020-09-19
Fix leftover references to the v1 API

giorio94:
# Description

This PR fixes some references which remained to the v1 API, instead of the v1alpha1 version as changed in #306 

# How Has This Been Tested?

Locally executing the frontend and the operator and verifying the correct functioning


palexster (approved):

GabriFila (approved):
lgtm

PR #308, opened 2020-09-20
Clean client code and improve UI and UX

GabriFila:
# Description

This PR is splitted in 2 main parts: 

### Part 1
Some cleaning and changes to the client code, mostly to UserLogic.jsx. This is done to reduce the complexity of the code and ease readability. This PR will be followed by others to clean the code, I prefer to avoid a single big PR to allow for better code reviewing and reduce bug passing.

Main changes:
- the info about the selected instance is no longer used, the data is provided by the calling function from the UI
- the UI doesn't use events so I removed the code, it is still on git in case we need it
- adminHidden was a bit confusing so I changed it to isStudentView

### Part 2

New features for better UI and UX

Main changes:
- adding the ability to filter the template and instances based on a text value
- adding a better name to vms and templates

### ***Important***
Do you think we should squash the commit before merging them? I added the `hold` label to wait for your answer

# How Has This Been Tested?

This has been tested by checking that the new names are correctly displayed and that the filter work. Also by checking that all the required functionalities still performed the same way as before:
- view templates and instances running
- start an instance
- connect to an instance
- stop an instance
- create a template
- delete a template

### DIFF
--- a/webservice/src/UserLogic.jsx
+++ b/webservice/src/UserLogic.jsx
@@ -16,7 +16,7 @@ export default class UserLogic extends React.Component {
    * - all ADMIN lab templates as a Map: (course_group => Array of available templates for that course)
    * - all ADMIN lab instances as a Map: (instance_name => URL if running, null otherwise)
    * - boolean variable whether to show the status info area
-   * - adminHidden whether to render or not the admin page (changed by the button in the StudentView IF adminGroups is not false)
+   * - isStudentView whether to render or not the admin page (changed by the button in the StudentView IF adminGroups is not false)
    * */
 
   constructor(props) {
@@ -60,7 +60,7 @@ export default class UserLogic extends React.Component {
       templateLabsAdmin: new Map(),
       instanceLabsAdmin: new Map(),
       adminGroups,
-      adminHidden: true
+      isStudentView: true
     };
     this.retriveImageList();
     this.retrieveCRDtemplates();
@@ -436,7 +436,7 @@ export default class UserLogic extends React.Component {
     const { logout } = this.props;
     const {
       name,
-      adminHidden,
+      isStudentView,
       adminGroups,
       registryName,
       imageList,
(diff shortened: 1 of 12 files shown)


webservice/src/UserLogic.jsx
```diff
@@ -284,34 +223,29 @@ export default class UserLogic extends React.Component {
...
+    if (!instanceLabsAdmin.has(instanceName)) {
+      Toastr.info(`The \`${instanceName}\` lab is not running`);
       return;
     }
     this.apiManager
       .deleteCRDinstance(
-        selectedInstance,
-        instanceLabsAdmin.get(selectedInstance).studNamespace
+        instanceName,
+        instanceLabsAdmin.get(instanceName).studNamespace
```
giorio94, line 237: Here the `studNamespace` property is used, but it is no longer set. Hence, the deletion of a laboratory from the professor view fails.

giorio94 (changes requested):
Great @GabriFila! The new names are much more user-friendly

Just a couple of issues I discovered while testing the UI:
1. The list of Running Labs disappears if the page is reloaded
2. The professor view does no longer display the owner the VM
3. It is not possible to delete VMs from the professor view (see the specific comment)

From my point of view, multiple commits are perfectly fine, since each one addresses a separate issue.

s41m0n (approved):

GabriFila:
> Great @GabriFila! The new names are much more user-friendly
> 
> Just a couple of issues I discovered while testing the UI:
> 
> 1. The list of Running Labs disappears if the page is reloaded
> 2. The professor view does no longer display the owner the VM
> 3. It is not possible to delete VMs from the professor view (see the specific comment)
> 
> From my point of view, multiple commits are perfectly fine, since each one addresses a separate issue.

Thanks for the review, regarding your points:
1. the list disappears for an issue related to the k8s watcher and some browser, in the future it could be fixed by not using the watcher if it fails
2. Fixed it in the last commit, also added capability to sort and filter by user
3. You are right, now it should be fixed

giorio94 (approved):
Thanks @GabriFila 

/approve

PR #312, opened 2020-10-01
Added visual discrimination between dekstop and CLI VMs

GabriFila:
# Description

This PR includes:
- added visual discrimination between VM type in the lists
- added possibility to create new template with custom description and type
- removed `labId` field from the newTemplateForm and made it self-generated, it was a bit confusing
- added template name to GUI, to make it more reachable with kubectl
- change parameters when creating new templates to make it same as python scripts
- fix visualization of creation date and labcode in instance view

This PR will remain in draft till the operator can create CLI only VMs, just to be sure that everything will work correctly

# How Has This Been Tested?

This has been tested by checking that all the required GUI functionalities work

- view templates and instances running
- start an instance
- connect to an instance
- stop an instance
- create a template
- delete a template

### DIFF
--- a/webservice/package.json
+++ b/webservice/package.json
@@ -103,6 +103,14 @@
       "react/jsx-pascal-case": "error",
       "react/prop-types": "off",
       "no-nested-ternary": "off",
+      "no-console": [
+        "warn",
+        {
+          "allow": [
+            "error"
+          ]
+        }
+      ],
       "react/jsx-indent": "off",
       "react/jsx-wrap-multilines": [
         "error",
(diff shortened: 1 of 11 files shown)


giorio94 (approved):
Thank @GabriFila 

Look perfect to me

palexster (approved):

GabriFila:
The last push was to make the created date in the local timezone, at [this line](https://github.com/netgroup-polito/CrownLabs/pull/312/files#diff-9700efb3b20e2afcb85eebcdbdb18ba0R271)

GabriFila:
Last push added check for user preferred color scheme before setting the theme for the first time

PR #313, opened 2020-10-04
Labinstance operator: add support for CLI VMs

giorio94:
# Description

This PR modifies the labinstance operator to add the support for CLI only VMs (i.e. detect when they are running).

### Additional modifications
* Changed the base docker image to `busybox`, for easier debugging
* Uniformed the name of the resources created by the operator (i.e. equal to the labinstance name)

# How Has This Been Tested?

Deploying the operator and verifying the correct functioning (both with GUI and CLI VMs).

### DIFF
--- a/operators/labInstance-operator/controllers/labinstance_controller.go
+++ b/operators/labInstance-operator/controllers/labinstance_controller.go
@@ -18,7 +18,7 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"net/http"
+	"net"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


palexster (approved):
/lgtm

operators/labInstance-operator/controllers/labinstance_controller.go
```diff
@@ -122,7 +127,7 @@ func (r *LabInstanceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	}
 
 	// prepare variables common to all resources
-	name := "l-" + labTemplate.Name + "-" + fmt.Sprintf("%.8s", uuid.New().String())
+	name := strings.ReplaceAll(labInstance.Name, ".", "-")
```
palexster, line 130: Can we get rid of uuid?

giorio94: I am honestly a bit uncertain about this.

If we consider the labinstances created by the front end, I feel there should be no problem since they are already characterized by a random number (I removed it to avoid having too many sources of randomness). 

On the other hand, if the labinstances are supposed to be created manually, then it may be better to keep it to avoid problems if one is created, removed and then shortly recreated (before the corresponding VM got deleted).

What is your opinion?

giorio94: In the end, I added back the UUID, just to be on the safe side

mLavacca (approved):

PR #314, opened 2020-10-08
Fix check to show switch view button only when admin

GabriFila:
# Description

This PR fixes a bug inserted by an [old PR](https://github.com/netgroup-polito/CrownLabs/pull/311/files#diff-87b3e833f5d97db4731f5f31c506c92eL85) of mine. Don't know why but I removed the check to show the *switch view button* only when the user is also admin.

I also changed to tooltip text depending on the current view

# How Has This Been Tested?

It has been tested by logging with 2 different users and checking the presence of the button.

### DIFF
--- a/webservice/src/components/Header.jsx
+++ b/webservice/src/components/Header.jsx
@@ -40,8 +40,10 @@ export default function Header(props) {
     name,
     logout,
     setIsLightTheme,
-    isLightTheme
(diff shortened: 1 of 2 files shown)


giorio94 (approved):
/lgtm

frisso (approved):

PR #315, opened 2020-10-10
Improve the Github Action pipelines

giorio94:
# Description

This PR improves the GitHub Actions pipelines:
* Add the support for PRs from forks
* Uniform the style between the different pipelines
* Remove some bugs that could happen in certain scenarios 

@GabriFila I removed the "changed paths" check from the website lint, since now it does not run too long (as it is executed in parallel with the others) and that check added only additional complexity (it is not possible to skip the entire job but only the single steps). Is that ok for you?

# How Has This Been Tested?

- [x] Checking that this PR triggers the correct checks

Unfortunately, the test regarding a PR from a fork cannot be executed until this one is merged.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Laboratory Operator - Check linting
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.31
+          working-directory: operators/labInstance-operator
+          args: --timeout=600s --enable gosec,prealloc,gofmt,govet --skip-files "zz_generated.*.go" --skip-dirs "kubeVirt"
+
+
+  javascript-lint:
```
GabriFila, line 28: I would call it `front-end-lint`

GabriFila:
The PR is ok to me, the filter is not necessary.

I just added a note, which can be also not changed

giorio94:
The last commit is a proposal of refactoring of the kubernetes manifests of the frontend and the laboratory operator.

In this refactoring, the manifests are no longer based on kustomize, since in my opinion it has some limitations that cannot be easily overcome. Most importantly (especially considering the version included in kubectl), the impossibility of easily configuring parameters at run time - such as the version of the docker image - and the need for large rewritings to modify simple fields, such as the hostname of an ingress controller. This, in turn, leads to a difficult interpretation of the final manifests (at least IMO), since it is hard to understand which parts of the original manifests (often not updated) are used and the ones that are overwritten.

I feel this alternative based on environment variables to be easier to understand (while maintaining the same production/preproduction workflow), although still not optimal. At the same time, Helm is probably overkill for the simple tasks to be performed. I am obviously open for alternatives.

*Note 1: this is only a proposal, and the commit can be easily dropped or moved to a different PR is this approach does not seem to be appropriate.*

*Note 2: a companion PR in CrownOps adapts the workflows using this manifests, while deploying tagged versions instead of latest and stable, which led to confusion and bugs.*

GabriFila (approved):
lgtm

mLavacca (approved):

palexster (approved):

PR #317, opened 2020-10-11
Uniforming ordering controls in GUI lists

GabriFila:
# Description

This PR includes:
- little fix in UX: the list Vm instances didn't actually remember the user's last choice due to a misalignment in the local storage key name
- added filtering based on VM type in template list, can be useful for uses who have access to multiple templates

*Note: this changes will make the transformation into a single list component (which will be duplicated through props) much easier*

# How Has This Been Tested?

This PR has been tested by checking that the order remains when reloading the page. It has been also checked that the changing of the order in the student view does not affect the one in the prof view


giorio94 (approved):
/lgtm

frisso (approved):

PR #342, opened 2020-11-09
lab-operator: fix bugs in the k8s manifest

giorio94:
# Description

This PR fixes a couple of problems related to the k8s manifest of the lab-operator:
* Adds a missing label to allow contacting the VMs from the operator
* Increases the memory limits, to prevent the operator from being   OOMkilled when started and many VMs are already present
* ~~Removes the readOnly protection, which caused the pod to crash due to  the logger trying to write files upon error~~ (keeping for now, as required by the kubescore checks)


palexster (approved):

GabriFila (approved):
/lgtm

PR #345, opened 2020-11-13
VM Provisioning: explicit Docker MTU for k8s

giorio94:
# Description

This PR explicitly sets the Docker MTU in the k8s role. This is necessary since the MTU in CrownLabs is lower than 1500 and Docker does not detect it automatically. The MTU is explicitly not configured for the Docker lab, in order to show how to solve this problem during the lab itself.


frisso:
@giorio94 Good catch: thanks!

frisso (approved):

giorio94:
> @giorio94 Good catch: thanks!

To be fair, it's @palexster's merit

palexster (approved):

Issue #210, opened 2020-04-20
[New feature] Container-based lab-instances


ReddaHawk:
# Problem
Currently our laboratories runs in VMs. Each virtual machine requires its underlying operating system, and then the hardware is virtualized. VMs, however, can take up a lot of system resources. Each VM runs not just a full copy of an operating system, but a virtual copy of all the hardware that the operating system needs to run. This quickly adds up to a lot of RAM and CPU cycles and implies low scalability. 

# Possible Solutions
A possible solution is to use Containers.
With containers, instead of virtualizing the underlying computer like a virtual machine (VM), just the OS is virtualized. Docker containers are easily portable and less resource-intensive than virtual machines and this is what we need in our dynamic environment. When Docker runs a container, it runs an image inside it. This image is usually built by executing Docker instructions, which add layers on top of existing image or OS distribution. 

We can use Containers to follow two different flavors:
1. Build a full image with a Desktop Environment and allow users to use it like we are doing now with vms. 
So we can build our laboratories from a same base layer and this allow us to use less storage in the registry.
This picture below show's performances consumed by a running container in my laptop that runs:
Ubuntu 18.04 with lxde DE and novnc setup. As you can see in idle uses 300 MB of memory and with firefox uses around 470MB.
![image](https://user-images.githubusercontent.com/44225903/79745024-b00eef80-8307-11ea-8a3e-159fba8672ea.png)
If you want try yourself you can run this command
```docker run -p 6080:80 -v /dev/shm:/dev/shm dorowu/ubuntu-desktop-lxde-vnc``` and when container starts browse http://127.0.0.1:6080/
We can also change DE and use for example XFCE4 or share audio only if users runs linux distos. For more information you can see [here](https://github.com/fcwu/docker-ubuntu-vnc-desktop)
Is the best one in my opinion.

2. Instead of building an entire image with Desktop Environment we can use containers to deploy the specific application for the specific laboratory. An example is to run GNS3 in a container like done [here](https://www.digiampietro.com/2014/1109/cisco_network_emulation_gns3_docker_container.html). With this method we will have some performance improvement,but we lose functionality: 
Crownlabs will be not a desktop exposer but only an application exposer, so will lose a lot of great stuffs done.
We must study how to install the specific application that we will want to expose and all the minimum dependencies required to work.
How we can share not only a single window of the application but also the pop up windows like terminals of the various routers in GNS3 trough novnc?

# Potential Benefits

- Improved Scalability
- Faster boot
[...]

Issue #94, opened 2020-03-25
Persistent disk improvements


frisso:
- [x] Currently, the content of the persistent folder on VMs is lost when we stop and restart the machine. The content of this folder should persist across VM reboots.

- [x] The content of the persistent folder is not shared among VMs of different labs. This folder should be a property of the user, hence should be mounted on all VMs belonging to the same user.

PR #347, opened 2020-11-17
Refactor LabTemplate API to support new scenarios

palexster:
# Description

This PR refactors the LabTemplate API to enable possible support to:

- [x] Persistent VMs (Ref. #94)
- [x] Container-based Laboratories (Ref. #210)
- [x] Support for multiple Lab-Environements (e.g. labTemplate which spawns multiple VMs)
- [x] DeleteAfter Condition

To do so, we are required to change the APIs:

- [x] Kubevirt objects are not embedded in the API anymore

# How Has This Been Tested?


operators/api/v1alpha2/labtemplate_types.go
```diff
@@ -0,0 +1,91 @@
...
+	Description     string        `json:"description,omitempty"`
+	EnvironmentList []Environment `json:"environmentList,omitempty"`
+	// +kubebuilder:validation:Pattern="^[0-9]+[mhd]$"
+	DeleteAfter string `json:"deleteAfter,omitempty"`
+}
+
+// TemplateStatus defines the observed state of Template
+type TemplateStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make lab-template" to regenerate code after modifying this file
```
giorio94, line 46:
```suggestion

```

operators/api/v1alpha2/labtemplate_types.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type Environment struct {
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Enum="GUI";"CLI"
+	LabType   `json:"labType,omitempty"`
```
giorio94, line 52: I feel it is hard to distinguish between `LabType` and `EnvironmentType`

What about replacing LabType with a boolean value like HasGui/NeedsGui/CLIOnly as done for Persistent?

operators/api/v1alpha2/labtemplate_types.go
```diff
@@ -0,0 +1,91 @@
...
+	Resources EnvironmentResources `json:"resources,omitempty"`
+	// +kubebuilder:validation:Enum="VirtualMachine";"Container"
+	EnvironmentType EnvironmentType `json:"environmentType"`
+	Persistent      bool            `json:"persistent"`
+	Image           string          `json:"image"`
+}
+
+type EnvironmentResources struct {
+	CPUNumber   uint32            `json:"cpuNumber,omitempty"`
```
giorio94, line 61: Do we want to consider only integer numbers or allow also things like 500m?
Probably kubevirt requires it to be integer, but we could round up in that case

Second, I would add an upper bound to avoid excessive values

Third, probably these fields should not have the omitempty annotation, since they are mandatory

operators/pkg/instanceCreation/creation.go
```diff
@@ -22,18 +22,78 @@ import (
...
+			Domain: virtv1.DomainSpec{
+				Resources: virtv1.ResourceRequirements{
+					Requests: v1.ResourceList{
+						"cpu":    template.Resources.ReservedCPU,
+						"memory": template.Resources.Memory,
+					},
+					Limits: v1.ResourceList{
+						"cpu":    template.Resources.ReservedCPU,
+						"memory": template.Resources.Memory,
+					},
```
giorio94, line 46: This should take into account the cpunumber and not ReservedCPU.

In addition, you should add some extra room for the hypervisor as we are doing manually now.

operators/api/v1alpha2/labtemplate_types.go
palexster: I left "omitempty" because it is not required or required for containers. Totally agree for having a upper bound.

operators/api/v1alpha2/common.go
```diff
@@ -0,0 +1,6 @@
+package v1alpha2
+
+type GenericRef struct {
```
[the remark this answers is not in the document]
palexster: types.NamespacedName is ok only for namespaced resources: Tenant are for example not Namespaced. No particular opinion on this, we can leave them expressed as two different structs.

operators/api/v1alpha2/labtemplate_types.go
giorio94: Ok, I understand.
I would make it nonetheless mandatory in this API to force the requests to be specified, but it is just an opinion

operators/api/v1alpha2/common.go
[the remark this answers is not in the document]
giorio94: Me neither, as you prefer

giorio94 (approved):

mlavacca (approved):

PR #349, opened 2020-11-23
Added types for tenant and workspace

GabriFila:
# Description

This PR adds basic types for the workspace and tenant CRDs.

I also added a `crd-example` folder where developers can have a quick view on how a yml file for the specific CR should be constructed. Tell me if it can be useful or it just makes things more confusing?

# How Has This Been Tested?

This PR has not official tests, they will be added when the controllers are added


operators/api/v1alpha1/tenant_types.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// TenantSpec defines the desired state of Tenant
+type TenantSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Name string `json:"name"`
+
+	SurName string `json:"surname"`
```
giorio94, line 51:
```suggestion
	Surname string `json:"surname"`
```

Just a proposal though

operators/api/v1alpha1/tenant_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// +kubebuilder:validation:MinLength=8
+	ID string `json:"ID"`
```
giorio94, line 54: Here, I would lower the minimum length, to avoid problems with students IDs

operators/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,20 @@
+package v1alpha1
+
+// NameCreated contains info about the status of a resource
+type NameCreated struct {
```
giorio94, line 4: Not clear to me. Should this be `NamespaceStatus` or is it meant to be more general?

operators/api/v1alpha1/tenant_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+// TenantStatus defines the observed state of Tenant
+type TenantStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	PersonalNamespace NameCreated `json:"personalNamespace,omitempty"`
+	SandboxNamespace  NameCreated `json:"sandboxNamespace,omitempty"`
+
+	Subscriptions map[string]SubscriptionStatus `json:"subscription,omitempty"`
```
giorio94, line 78: Would it be possible to replace the string with an enum?

giorio94 (commented):
Thanks @GabriFila, just a couple of minor comments.

As for the examples, they may definitely be useful (I would just put them in `deploy/samples`). The only problem is that they will soon be out of date, but we can manage this problem in the future with some automated check.

giorio94 (approved):
/lgtm

palexster (approved):

PR #352, opened 2020-11-24
Add ssh access to VMs

claudious96:
# Description
This PR presents a possible implementation of an SSH bastion using a sidecar approach. Each pod is composed of an ssh-bastion running sshd and a sidecar operator that whatches the Tenant resource updating the `authorized_keys` file at each tenant related event (creation/deletion/update of a Tenant resource).

Currently the operator gets a `List` of Tenants in the default namespace, this is good for testing but probably should be managed differently.
Assuming that a single deployment would manage all incoming ssh incoming traffic for the whole crownlabs cluster, since the Tenant resource could belong to one or more workspaces, a possible approach could be:
- establish (somehow) a list of workspaces that are going to use this feature i.e. Cloud Computing course
- perform a List operation on the Tenant CRD belonging to those workspaces
- update the `authorized_keys` file

Otherwise if the plan is to deploy it per workspace, the operator could just perform the `List` on the related workspace and the operator should be modified in order to be aware of the workspace (namespace might be useful).

Notes:

- [x] bastion's host keys are generated in the Dockerfile but they should probably be set using a secret
- [x] sshd configuration is created using a RUN command, could be created as file to be copied at build time
- [x] give the possibility to set parameters such as port number, namespace name, etc (just like it's done for the lab operator)
- [x] default message in /etc/motd/ in case a user connects to the bastion without the -J jump option
- [x] Ginkgo tests


frisso:
Quoting from the PR description:
> Resource updating the authorized_keys file at each tenant creation/deletion.
I would suggest to update the `authorized_keys` also when the tenant is updated. It would be useful to allow users to update the list of authorized keys in their profile.

operators/pkg/controllers/bastion_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+// +kubebuilder:rbac:groups=crownlabs.polito.it,resources=tenants,verbs=list
+
+func (r *BastionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("bastion", req.NamespacedName)
+	log.Info("reconciling bastion")
+
+	// Get tenants resources
+	var list crownlabsalpha1.TenantList
```
giorio94, line 47:
```suggestion
	var tenants crownlabsalpha1.TenantList
```

Just a proposal, to make the code more readable

giorio94 (commented):
Thanks @claudious96. At a first sight, this PR seems very good!

A couple of comments with regard to your notes (and a few very minor inline):
1. Tenant resources are not namespaced 
   https://github.com/netgroup-polito/CrownLabs/blob/a2b3c249ce7f1e5e1b3267fc61425f4e30a0c4cf/operators/deploy/crds/crownlabs.polito.it_tenants.yaml#L17
2. Totally agree about using the keys from a secret and not statically generate them in the docker
3. From my point of view, both generating the configuration using RUN and using an external file is fine. In the first case, you could use only one echo instruction instead of multiples.
4. Fine to have a separate manifest with the cluster role (makes it easier to apply the others from the CI)
5. Yes, it would be nice to have a bit of templating for the namespace, image version and so on

claudious96:
Thanks for the input @frisso , I've just tested the case (updated the keys on an existing Tenant resource) and the `authorized_keys` file is correctly updated.
I'll update the description.

claudious96:
@giorio94 I updated the PR with your suggestions. 

The manifest can now be generated like the lab operator, setting desired values, then exporting the envs and substituting them in the template. In case of using the default namespace the result of the kubectl apply is `namespace/default configured`, I'm not sure if actually changes something though.

I wrote some instructions in the README inside the `infrastructure/ssh-bastion` folder to generate host keys and create the relative secret. I'm not sure if it's okay to have the inscructions there since the kubectl command depends on the namespace the deployment is going to run.

I have also updated the var you suggested in the bastion controller and the message in `motd`.

operators/pkg/controllers/bastion_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+	f, err := os.Create("/auth-keys-vol/authorized_keys")
+	if err != nil {
+		log.Error(err, "unable to create the file authorized_keys")
+		return ctrl.Result{}, nil
+	}
+	_, err = f.WriteString(authorizedKeys)
+	if err != nil {
+		log.Error(err, "unable to write to authorized_keys")
+		f.Close()
```
giorio94, line 68: It is probably better to use `defer closeFile()` right below when you open the file, to ensure it is always closed (where closeFile() is a function which closes the file and checks for errors as you are doing below)

operators/deploy/bastion-operator/k8s-cluster-role.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: crownlabs-bastion-operator
+rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["get","list","watch"]
+
+- apiGroups: [""]
+  resources: ["secrets","services","events"]
+  verbs: ["get","list","watch","create"]
```
giorio94, line 12: Are these permissions required?

operators/deploy/bastion-operator/k8s-manifest.yaml.tmpl
```diff
@@ -0,0 +1,108 @@
...
+            mountPath: /auth-keys-vol
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 20000
+          runAsGroup: 20000
+          privileged: false
+      - name: bastion
+        args: ["-D", "-e", "-f","/etc/ssh/sshd_config_custom"]
+        command: ["/usr/sbin/sshd"]
+        image: crownlabs/ssh-bastion${IMAGE_SUFFIX}:${IMAGE_TAG}
```
giorio94, line 69: I would use a different IMAGE_TAG from the one above, as this is an infrastructural component which may be not be regenerated every time

giorio94 (approved):
/lgtm

Just a couple of minor comments inline. 
Could you also add the `bastion-operator` to this list in order to automatically generate the Docker images? https://github.com/netgroup-polito/CrownLabs/blob/31092a7aff5e0055a10d83e45e651848dd8c0b83/.github/workflows/build.yml#L91-L94

palexster (approved):

PR #355, opened 2020-11-26
Fix the code testing and linting pipelines

giorio94:
# Description

This PR fixes a bug introduced in #351. Specifically, since the pipeline was changed to run on the `pull_request_target` event, the checkout action started getting the master branch instead of the last commit of the PR.

Additionally, it also changes the linting pipeline to use pull_request_target (as it requires the PAT) and applies the same modifications.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,6 @@
 name: Check linting
 on:
-  pull_request:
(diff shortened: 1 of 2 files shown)


palexster (approved):

GabriFila (approved):
lgtm

PR #356, opened 2020-11-26
Add TigerVNC and noVNC Dockerfiles

qcfe:
# Description

This PR adds two Dockerfiles for generating containers containing the following: 
- VNC server (TigerVNC + fluxbox)
- noVNC server (nginx + noVNC dist + websockify-c)


provisioning/containers/tigervnc/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu:20.04
+
+ARG USER=vncuser
```
giorio94, line 3:
```suggestion
ARG USER=crownlabs
```

If the name doesn't matter, I would go with crownlabs

provisioning/containers/tigervnc/fluxbox/politheme
```diff
@@ -0,0 +1,189 @@
+#This work is licensed under a Creative Commons Attribution-Share Alike 3.0 License.
+#(http://creativecommons.org/licenses/by-sa/3.0/de/deed.en)
+
+--------------------
+
+background:                             aspect
+background.color:                       #003576
+background.colorTo:                     #003576
+
+background.pixmap:      ~/.fluxbox/crownexam.png
```
giorio94, line 10: Nice picture :). I would just change the filename in ~/.fluxbox/wallpaper.png or something like that

provisioning/containers/novnc/nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+
+    #gzip  on;
+
+
+      upstream websockify {
+          server 127.0.0.1:8888;
+      }
+
+    server {
+      listen       80;
```
giorio94, line 34: I would use a non standard port here (e.g. 8080), to simplify running the container as non-privileged

provisioning/containers/novnc/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Configure default VNC endpoint
+ENV VNC_SERVER=localhost \
+    VNC_PORT=5900
+
+# Start ngix not in daemon mode and websockify
+# then wait in order to stop in case any of the two crashes
+CMD nginx -g "daemon off;" &\
+    websockify 8888 $VNC_SERVER:$VNC_PORT &\
+    wait -n
```
giorio94, line 44: Wondering whether to split this in two separated containers, each with its own entrypoint (to keep one process per container). What do you think?

provisioning/containers/novnc/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM nginx:alpine
```
giorio94, line 1: I would explicitly specify an nginx version, to avoid incompatibility issues.

giorio94 (commented):
Thanks @qfce. Seems really really good to me. 

I just left a couple of minor comments inline.

provisioning/containers/tigervnc/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    apt-get install -y \
+        fluxbox \
+        tigervnc-standalone-server && \
+    apt-get --purge autoremove -y && \
+    apt-get autoclean -y
+
+EXPOSE 5900
+
+### Prepare usermode
+RUN useradd -ms /bin/bash $USER
```
giorio94, line 21: Here, I would explicitly configure a uid, since it is necessary to specify it in the k8s deployment.

provisioning/containers/novnc/Dockerfile
qcfe: Yes, that would make sense, though the idea here was to "hide" websockify entrypoint behind the nginx proxy. If it's not an issue to have 3 sidecars anyway for sure it's not an issue to split this!

provisioning/containers/tigervnc/Dockerfile
qcfe: which uid has to be specified? 1000

provisioning/containers/novnc/Dockerfile
giorio94: If I understand correctly, websockify would be anyhow behind the proxy, just in a different container (as pods share the same network namespace, you can contact it on localhost). From my point of view, it is definitely not a problem having one more sidecar (and the process management is easier).

provisioning/containers/novnc/nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+          proxy_set_header Connection "Upgrade";
+          proxy_set_header Host $host;
+      }
+      
+      location / {
+          add_header Cache-Control no-cache;
+          root   /usr/share/nginx/html;
+          index  vnc.html;
+      }
+
```
giorio94, line 50: I would suggest to add a healthz location, to enable the addition of a readiness probe

```
location /healthz {
        access_log off;
        return 200 "healthy\n";
}

```

giorio94 (commented):
Forgot earlier

provisioning/containers/tigervnc/Dockerfile
giorio94: Really not sure about this. I would suggest 2000 to avoid matching the uid on the host, but I do not know if there is any "best practice"

@palexster what is your opinion?

provisioning/containers/tigervnc/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+ENV HOME=/home/$USER
+
+# Copy fluxbox config & vnc session tweaks
+COPY --chown=${USER}:${USER} fluxbox $HOME/.fluxbox
+COPY Xvnc-session /etc/X11/Xvnc-session
+
+CMD rm -rfv /tmp/.X11-unix/* && \
+    su -p $USER -c "\
```
giorio94, line 29: Backporting from #357 

su should not be necessary if you use the USER command in the dockerfile (docs.docker.com/engine/reference/builder/#user)

provisioning/containers/tigervnc/Dockerfile
qcfe: Do I have to specify uid and gid to match a similar pattern?
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

giorio94: Exactly, the idea is to configure the uid you choose in the security-context of the pod

qcfe: So, I checked in the only container that has the user already set, that is the nginx one. The uid there is 101. 
I was wondering for praticity if we can set 101 as the default uid in all the other containers at this point...

giorio94: I would avoid assigning a uid < 1000 to a "human" user as it is typically used for services, even though this distinction probably is not that important in k8s

qcfe: Ok but at this point I have to change the uid for the nginx user (in the nginx container) I guess?

giorio94: As it is possible to [specify a different securityContext for each container](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container) in a pod, that should be not necessary

qcfe: I added a new user on the nginx container and it seems to work, I'm pushing the changes now

giorio94 (approved):
/lgtm

I would just propose to group the three folders in a parent one (e.g. `common/`) to clarify that these are common components to all applications if @palexster also agrees.

Finally, could you squash the two commits and rebase to remove the merge commit? Thanks!

palexster (approved):

PR #372, opened 2020-12-13
Fix and improve workspace controller

GabriFila:
# Description

This PR includes:
- fix to deployment files for the workspace CRD controller
- logic flow fix in function to create a role in keycloak
- adapted the workspace controller to use a `fail-fast:false` strategy, in order to have a cleaner code and to avoid executing some actions if other independent ones have failed.

_Side note:_ please tell me what you think about the use of a dedicated struct for the retriggering of the reconcile, I believe it is an elegant solution. If it is ok, il later PRs I would move it to another file, since it will be used also by the tenant-controller.

# How Has This Been Tested?

This has been tested by temporarily adding an error in the code and check the correct retrigger of the reconcile function. Also, since the injected error consisted in providing a non-existing name for the target realm for the creation of keycloak roles, this procedure allowed to spot a bug in the logic of  `createKcRole` in the `keycloak_utils.go` file. The bug made the operator crash.

### DIFF
--- a/operators/api/v1alpha1/common_types.go
+++ b/operators/api/v1alpha1/common_types.go
@@ -7,14 +7,12 @@ type NameCreated struct {
 }
 
 // SubscriptionStatus is an enum for the status of a subscription to a service
-// +kubebuilder:validation:Enum=Ok;Pending;Failed
+// +kubebuilder:validation:Enum=Ok;Failed
 type SubscriptionStatus string
 
 const (
 	// SubscrOk -> the subscription was successful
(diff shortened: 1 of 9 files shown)


GabriFila:
Thanks @giorio94 for the suggestion about the solution on the `fail-fast:false` strategy. I highlight the fact that at this point, the value of `pending` for the status of a subscription doesn't apply, since it will always be `failed` or `ok`

giorio94 (approved):
/lgtm

palexster (approved):

PR #377, opened 2020-12-15
SSH access: improve bastion-operator

claudious96:
# Description

This PR aims at improving the bastion-operator reducing the load on the K8s api server. Instead of listing all the tenants each time that a CRUD event happen, it only gets the related tenant resource and recreates the `authorized_keys` file.

Here is how:
1. Each entry on the file has now this syntax>: `<algorithm> <key> <comment>`
2. In place of the comment, the Tenant's name/id is inserted. This way it is possible to associate a key to its owner.
3. At each event all the keys related to that tenant will be removed (if any)
4. The updated keys will be appended to the list
5. If the tenant has been deleted, the keys will only be removed without further actions

At each reconciliation the file is being read and the written again from scratch, so that there are no possibilities of duplicated entries.

# How Has This Been Tested?
BDD tests with Ginkgo

## Notes:
- [x] tests could be more tidy
- [x] move helper functions used in the bastion controller in a separated file


operators/pkg/bastion-controller/bastion_controller.go
```diff
@@ -37,35 +38,41 @@ type BastionReconciler struct {
...
-	}
+	tenant := &crownlabsalpha1.Tenant{}
+	deleted := false
 
-	// Collect public keys of tenants
-	keys := make([]string, 0, len(tenants.Items))
-	for _, tenant := range tenants.Items {
-		keys = append(keys, tenant.Spec.PublicKeys...)
+	if err := r.Get(ctx, req.NamespacedName, tenant); err != nil {
+		deleted = true
```
giorio94, line 52: I would add a check that the error is actually NotExists, and not a different kind of error

operators/pkg/bastion-controller/bastion_controller.go
```diff
@@ -37,35 +38,41 @@ type BastionReconciler struct {
...
 
-	authorizedKeys := strings.Join(keys[:], "\n")
+	var keys []string
+
+	if _, err := os.Stat(r.AuthorizedKeysPath); err == nil {
 
-	log.Info(r.AuthorizedKeysPath)
+		// if the file exists, read the whole file in a []byte
+		data, err := ioutil.ReadFile(r.AuthorizedKeysPath)
+		stringData := string(data)
```
giorio94, line 61: Should be moved after the error check

operators/pkg/bastion-controller/bastion_controller_test.go
```diff
@@ -15,67 +15,103 @@ import (
...
-
-		NameTenant2          = "s22222"
-		NamespaceTenant2     = ""
-		FirstNameTenant2     = "Fabio"
-		LastNameTenant2      = "Bianchi"
-		IDTenant2            = "s22222"
-		EmailTenant2         = "fabio.bianchi@fakemail.com"
-		CreateSandboxTenant2 = true
+		NameTenant1      = "s11111"
+		NamespaceTenant1 = ""
```
giorio94, line 22: Could be dropped (the same for tenant2).

giorio94 (approved):
/lgtm

Just a couple of minor comments inline

claudious96:
Thanks @giorio94 for the review. I have updated the code according to your suggestions.

After the correction, the first test ran in the CI failed, I think because of the default timeout of the `Eventually` function used inside the `BeforeEach`, updating timeout and interval seems to get it right. However it could be a flaky test, could you trigger another run? On my machine the test is always successful and was also before updating the timeout.

Moreover, handling the deletion case (checking if the error was a NotFound) added another case that is quite difficult to test (I'm not really sure how to generate another kind of error on a GET on a tenant resource).
This results in a decrease of code coverage, and that blocks the merging in any case now.

I could try adding more test unit-like covering some edge cases that are currently not covered in the helper functions, that should increase the code coverage and solve that issue.

I do not really know why the `crownlabs-image-list` build failed.

giorio94:
> Thanks @giorio94 for the review. I have updated the code according to your suggestions.

Thanks!
> 
> After the correction, the first run of the test in the CI failed, I think because of the default timeout of the `Eventually` function used inside the `BeforeEach`, updating timeout and interval seems to get it right. However it could be a flaky test, could you trigger another run? On my machine the test is always successful and was also before updating the timeout.

Re-executed and worked correctly. Probably the runner was particularly slow that time. 
> 
> Moreover, handling the deletion case (checking if the error was a NotFound) added another case that is quite difficult to test (I'm not really sure how to generate another kind of error on a GET on a tenant resource).
> This results in a decrease of code coverage, and that blocks the merging in any case now.

Don't worry about that. The coverage check is not mandatory, and it does not block the merge of the PR. Its main purpose is to get an overall idea about the coverage, but small decreases are clearly possible as not all cases can be easily tested (especially the ones associated with errors and failures hard to replicate).

> I could try adding more test unit-like covering some edge cases that are currently not covered in the helper functions, that should increase the code coverage and solve that issue.

Clearly, more tests are always welcome :-)
> 
> I do not really know why the `crownlabs-image-list` build failed.

It seems a temporary failure. Rettrigered and worked correctly

palexster (approved):

PR #378, opened 2020-12-16
Vm to docker/images

qcfe:
# Description

This PR includes 
- documentation for the containers infrastructure 
- a little change in the novnc container, that now shows the novnc toolbar by default (and can be hidden through an ENV flag)

### DIFF
--- a/provisioning/README.md
+++ b/provisioning/README.md
@@ -6,3 +6,7 @@ For more detailed information, please refer to the respective *README* files.
 ## Virtual Machines
 The [virtual-machines](virtual-machines/) folder contains the resources made available to automate the creation, configuration and export of virtual machines to be used in *CrownLabs*. Additionally, it provides a set of scripts to configure and export already existing VMs, as well as to interact with the docker registry.
 These scripts do not represent yet an optimal solution. Indeed, the operations still require some manual intervention from the administrators (e.g. admin credentials are required to interact with the *docker registry* to upload the resulting images).
+
+## Containers
+
+The [containers](containers/) folder contains the resources made available to create an application, running in a container, which is compatible with *CrownLabs*, e.g., that exports a remotely-accessible GUI via web browser. This folder will actually contain a set of containers running together (with the Kubernetes *sidecar* approach) in order to make the solution easier to maintain, e.g., splitting the creation of the virtual desktop environment from the actual application container.
--- a/provisioning/containers/README.md
+++ b/provisioning/containers/README.md
@@ -1,0 +1,23 @@
+# Creating and uploading Docker-based graphical applications in CrownLabs
+
+This folder contains the files required to create docker-based graphical applications in CrownLabs, as an alternative solution to VMs.
+
+In order to simplify the process of creating such GUI applications, multiple containers are envisioned, running according to the *sidecar* model in Kubernetes:
+
+- **Common components**: a first set of containers, available under [gui-common](gui-common/), providing all the components required to provide a virtual desktop accessible from a web browser (i.e., virtual desktop environment, noVNC server, etc).
+- **Containerized application**: a further container keeping the actual application, running in the virtual desktop above. Currently, the [pycharm](pycharm/) folder provides an example of such an application, based on the PyCharm integrated development environment for Python.
+
+This architecture decouples the application itself, e.g. PyCharm, from the virtual desktop logical layer provided by `gui-common`.
+
+
+## Starting the application
+The application can be started by running all the involved containers within the same pod.
+
+
+## Creating a new application
+A new application can be created by just *containerizing* it, i.e., writing a self-contained Dockerfile for the application layer, without introducing strict bindings with the graphical components that would make the deployment monolithic and thus more difficult to maintain.
+In other words, we can deploy new applications by creating new dockerfiles modeled after the [PyCharm](pycharm/) one, provided as example, while leaving untouched the content of the `gui-common` folder.
+
+To the sake of completeness, two information may be considered when creating the `Dockerfile` for a new application:
+- the `DISPLAY` environment variable must have the same value in the `X` server instance and in the application container. In the current implementation, this is achieved by the custom Kubernetes operator that is in charge of starting the pod, which sets the same value in both.
+- the `/tmp/.X11-unix` directory that will contain the socket file used to communicate with the X server must be shared between the application container and the X VNC server. In the current implementation, this is achieved by the custom Kubernetes operator that is in charge of starting the pod, which mounts an `emptyDir` volume under `/tmp` on both containers.
--- a/provisioning/containers/gui-common/README.md
+++ b/provisioning/containers/gui-common/README.md
@@ -1,0 +1,72 @@
+# General architecture
+
+This folder contains the components required to create the *common ground* for graphical containers, which are the following:
+
+- **X server**: it is the foundation that allows for graphic elements to be drawn on the display. X Windows builds the primitive framework that allows moving of windows, interactions with keyboard and mouse, and draws windows. This is required for any graphical desktop.
+- **Window Manager**: it is the component that controls the placement and appearance of windows, such as sub-windows or menu opened by the application. It requires X Windows but not a desktop environment.
+- **VNC server**: it is a graphical desktop-sharing system that allow to remotely control another computer. It transmits the keyboard and mouse events from one computer to another, relaying the graphical-screen updates back in the other direction, over a network.
+- **Web server for noVNC**: an HTML5 implmentation for a VNC client is available for modern browsers; the web server is required to transform the VNC server into a web-based protocol, which enables users using a simple web browser to connect and show the remote display.
+
+For the sake of completeness, graphical containers do not require a Desktop Environment component (e.g., KDE, GNOME, etc), which represent a far more integrated and complex system than a Window Manager. In fact, a Desktop Environment usually includes a suite of applications that are tightly integrated with the system (control panel, networking controls, general purpose applications), as well as graphical widgets such as the *start menu*, status bars, system trays (for notifications and different widgets), and more. For more information, look at this [page](https://www.ghacks.net/2008/12/09/get-to-know-linux-desktop-environment-vs-window-manager/).
+
+
+## Current implementation
+
+The current implementation leverages three containers, each one contained in the following sub-folders:
+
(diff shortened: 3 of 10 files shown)


provisioning/containers/gui-common/novnc/Dockerfile
```diff
@@ -56,4 +52,7 @@ EXPOSE ${HTTP_PORT}
 USER ${USER}
 
 # Start nginx (not in daemon mode for Docker use)
-CMD nginx -g "daemon off;"
+CMD if [ $HIDE_NOVNC_BAR = true ] \
+        then echo "#noVNC_control_bar_anchor {display:none !important;}" \
+        >> $HTML_DATA/app/styles/base.css; fi\
+    nginx -g "daemon off;"
```
giorio94, line 58: Here, I would personally move these instructions to a script file and execute that as entrypoint (personal opinion, let me know what do you think).

Second comment, you should probably quote "$HIDE_NOVNC_BAR" to avoid issues if the variable is not set.

giorio94 (approved):
/lgtm

Just a couple of very minor comments

frisso:
First, great for this commit!
I would like to suggest the following changes.

- [ ] Add a small paragraph in file `https://github.com/netgroup-polito/CrownLabs/tree/master/provisioning` in order to introduce what's in the `container` folder.

- [ ] Add a small README in the `https://github.com/netgroup-polito/CrownLabs/tree/master/provisioning/containers` folder in order to explain what's in there. Also say what the user has to do to create new applications in there (so, what it can be reused, what else has to be written from scratch).

- [ ] Add a README in the `https://github.com/netgroup-polito/CrownLabs/tree/master/provisioning/containers/gui-common` folder, explaining the overall architecture of the solution (which containers are being created, what they are used for, how do they interact). 

- [ ] In general, add a README to all the folders that you created. I don't want to repeat myself, but who comes here may not know what is "fluxbox", what's in there, what is this used for. Documentation is the gold key to achieve maintainability. 

- [ ] Document more the Dockerfile you wrote. It would be very nice if all the commands in there are preceded by a small comment saying what you're doing and (even more important) why is this command needed, if this is meaningful (do not write obvious things, otherwise people would not read anymore).

In case some pictures are needed, you can consider uploading either in PNG or (if are supported by github; not sure, honestly) in SVG, so that it would be easier to modify and update in the future. Remember that pictures are often much more useful than a lot of text.

Thanks!

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,22 @@
...
+- Websockify
+- noVNC client files (static HTML files hosted somewhere)
+- the GUI application
+
+## Current implementation
+
+Currently the following containers are used to build the overall GUI application:
+
+- Xvnc server (based on TigerVNC) + fluxbox window manager, which are responsible to provide the basic graphical desktop environment.
+- Websockify (C version), exposing VNC server as a web server
```
frisso, line 20: Document (maybe not here, but it has to be said) why you're using the C version.

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,22 @@
+# Docker-based GUI
+
+This sub-project aims to provide a Dockerized environment where Docker-based GUI applications can run. This is backed by a set of different containers, each with its own specific functions, that together constitute and offer the Docker-based GUI application.
+
+## General architecture
+
+In order to deploy GUI applications in CrownLabs the following is required:
+
+- A VNC-based X server
+- A window manager (it's better to have it since sub-windows or menu opened by the application might not get rendered correctly)
```
frisso, line 10:
```suggestion
- A window manager (required to allow sub-windows or menu opened by the application to be rendered correctly)
```

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,22 @@
+# Docker-based GUI
+
+This sub-project aims to provide a Dockerized environment where Docker-based GUI applications can run. This is backed by a set of different containers, each with its own specific functions, that together constitute and offer the Docker-based GUI application.
+
+## General architecture
+
+In order to deploy GUI applications in CrownLabs the following is required:
```
frisso, line 7: "the following" what?
Maybe "he following components are required"?

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+- A VNC-based X server
```
frisso, line 9: "A VNC-based X server"
or "An X11 server with VNC support"?

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,22 @@
...
+- A window manager (it's better to have it since sub-windows or menu opened by the application might not get rendered correctly)
+- Websockify
+- noVNC client files (static HTML files hosted somewhere)
+- the GUI application
+
```
frisso, line 14: Maybe not here, but you have to document how those components are installed and run. Everything in the same Docker? Or in multiple Dockers? (e.g., sidecars)?

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,22 @@
+# Docker-based GUI
+
+This sub-project aims to provide a Dockerized environment where Docker-based GUI applications can run. This is backed by a set of different containers, each with its own specific functions, that together constitute and offer the Docker-based GUI application.
```
frisso, line 3: This sentence looks ok, but not here. Probably is much better in the (parent) "containers" folder.

provisioning/README.md
```diff
@@ -11,3 +11,7 @@ The [courses](courses/) folder contains the resources made available to create *
 
 ## Virtual Machines
 The [virtual-machines](virtual-machines/) folder contains the resources made available to automate the creation, configuration and export of virtual machines to be used in *CrownLabs*. Additionally, it provides a set of scripts to configure and export already existing VMs, as well as to interact with the docker registry.
+
+## Containers
+
+The [containers](containers/) folder contains the resources made available to create a set of containers that will run together in order to create a virtual desktop environment on top of which a desktop GUI application can run and be accessed as a web application through a web browser.
```
frisso, line 17:
```suggestion
The [containers](containers/) folder contains the resources made available to create an application, running in a container, which is compatible with *CrownLabs*, e.g., that exports a remotely-accessible GUI via web browser. This folder will actually contains a set of containers running together (with the Kubernetes *sidecar* approach) in order to make the solution easier to maintain, e.g., splitting the creation of the virtual desktop environment from the actual application container.
```

provisioning/containers/README.md
```diff
@@ -0,0 +1,20 @@
...
+This architecture decouples the application itself, e.g. PyCharm, from the virtual desktop logical layer provided by _gui-common_.
+
+## Starting the application
+The application can be started by running both containers in the same pod.
+
+## Creating a new application
+A new application can be created by just *containerizing* it, i.e., writing a self-contained Dockerfile for the application layer, without introducing strict bindings with the desktop environments that would make the deployment monolithic and thus less maintainable.
+In other words, we can deploy new applications by replacing the PyCharm's Dockerfile with a new one, while keeping what is inside _gui-common_.
+
+Note that the `DISPLAY` environment variable must be the same used within the running `X` server instance and that both Xvnc and the application container must share the `/tmp/.X11-unix` directory that will contain the socket file used to communicate with the X server.
```
frisso, line 19: I am not sure the message is so clear here. Can you please try to rephrase, or at least comment here, so that we can improve the readability of the sentences?

qcfe: The note on lines 19 and 20 is just for saying that who will make the application Docker image might want to be aware of those two "parameters" (the `DISPLAY` variable and the folder that will be mounted on `/tmp`), that will be set by the CrownLabs operator when issuing the instance of such application, but actually I guess an application could work perfectly without knowing about that, so the note could also be removed... It's more of a warning, just in case

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,76 @@
...
+Technically, this consists in two different servers (the actual X server and a VNC server) that are bound together and provide a virtual desktop both to the user (through the VNC protocol) and to the final application (through the X core protocol).
+
+There are several implementations of such type of server. For CrownLabs, [TigerVNC](https://github.com/TigerVNC/tigervnc) has been chosen, since it provides automatic desktop rescaling features.
+Note that TigerVNC server cannot be considered a regular X server since it cannot display the desktop on a physical screen.
+
+This container also includes [Fluxbox](http://fluxbox.org/), a lightweight window manager which provides an effectively desktop-like user experience and, above all, it enables menus and sub-windows to be rendered correctly.
+
+Note that the X server will create a socket file inside `/tmp/.X11-unix/`. You may want to share this directory with any client of such X server, along with setting the same `DISPLAY` environment variable.
+
+TODO (I did not understand the previous sentence)
```
frisso, line 36: Please review this "todo"

provisioning/containers/gui-common/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+In the context of CrownLabs, this container can be used in two ways:
+- as a sidecar inside the final application pod running the VNC server
+- as a disjoint deployment, decoupled from the final application pod running the VNC server
+
+In the first case, this container is tightly bounded to the VNC server, hence creating a pod that directly accepts (multiple) websocket connections.
+
+In the second case, this container can be instantated _once_ in the entire datacenter, serving all the VNC-based containers. This results in a more efficient use of resources, but it requires the ingress controller to be able to multiplex the websocket connection towards the correct service.
+
+TODO (I DID NOT UNDERSTAND)
```
frisso, line 75: Please review this "todo"

frisso:
First, I apologize... I pushed on the "update branch" button by mistake :-(
Second, this PR looks rather good to me . Please:
- review the text that I wrote, just to make sure that I understood everything correctly
- please document file `nginx.conf.template`, whose directive do not look obvious to me
- I wonder if we have to document the 'USER' and 'UID' which are set in all Dockerfiles

Finally, I would suggest to transform the websockify Docker into a multi-stage build:
- you create a first docker that downloads the code, compiles it
- then, you create, from scratch, a second docker that takes the result from the former (i.e., the executable), copies into it, and terminates. So, the user can simply use the second Docker, discarding the first.
Thanks!

frisso (approved):

PR #384, opened 2020-12-27
SSH access: workspace managers key injection, update laboperator rbac

claudious96:
# Description

This PR updates the permission of Laboperator so that it is possible to retrieve the related tenant and workspace managers.
In addition it brings the feature of retrieving also the workspace managers keys in order to inject them in the VM.

It also fixes the RBAC of bastion-operator, since the `list` action is not performed anymore, the permission to `get` is granted instead.

# How Has This Been Tested?

At the moment no tests have been performed, apart from a successful build and run of the operator.


giorio94 (approved):
/lgtm

palexster (approved):

PR #391, opened 2021-01-02
Support for GUI-based containers in the instance operator

qcfe:
This PR includes the possibility for the instance operator to run graphical applications inside regular containers.
It uses a set of sidecars to obtain the necessary environment.


operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+func CreateGUIContainerPod(name string, namespace string, template *crownlabsv1alpha2.Environment, instanceName string, references []metav1.OwnerReference, examMode bool) (appsv1.Deployment, error) {
+
+	var examModeStr string
+	if examMode {
+		examModeStr = "true"
+	} else {
+		examModeStr = "false"
+	}
```
giorio94, line 26: Should be replaced with [FormatBool](https://golang.org/pkg/strconv/#FormatBool)

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+
+	tmpXVolumeMounts := corev1.VolumeMount{
+		Name:      "x-tmp-dir",
+		MountPath: "/tmp/.X11-unix",
+	}
+
+	containers := []corev1.Container{
+		{
+			Name:  "novnc",
+			Image: "qcfe/novnc",
```
giorio94, line 36: Should be changed to use the crownlabs images, with the tag configured as a parameter

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+			Ports: []corev1.ContainerPort{{
+				ContainerPort: 80,
+			}},
```
giorio94, line 39: Why is this port exposed?

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+	} else {
+		examModeStr = "false"
+	}
+
+	tmpXVolumeMounts := corev1.VolumeMount{
+		Name:      "x-tmp-dir",
+		MountPath: "/tmp/.X11-unix",
+	}
+
+	containers := []corev1.Container{
```
giorio94, line 33: Requests and limits are missing from the sidecar containers

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+			Name:            name + "-guicont-depl",
+			Namespace:       namespace,
+			OwnerReferences: references,
+			Labels: map[string]string{
+				"name":          name,
+				"template-name": template.Name,
+				"instance-name": instanceName,
+			},
+		},
+		Spec: appsv1.DeploymentSpec{
```
giorio94, line 99: It is missing the SecurityContext specification

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+				Spec: v1.PodSpec{
+					Containers: containers,
+				},
+			},
+		},
+	}, nil
+}
+
+func (r *LabInstanceReconciler) CreateContainerEnvironment(
+	labInstance *crownlabsv1alpha2.Instance,
```
giorio94, line 111:
```suggestion
	instance *crownlabsv1alpha2.Instance,
```

All references to the old `lab*` terminology should be avoided

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+			Kind:               labInstance.Kind,
+			Name:               labInstance.Name,
+			UID:                labInstance.UID,
+			BlockOwnerDeletion: &b,
+		},
+	}
+	ctx := context.TODO()
+
+	// create Service to expose the vm
+	service := instance_creation.ForgeService(name, namespace, globalOwnerReference)
```
giorio94, line 130: Duplicated code is code hard to maintain and that will soon start to diverge, introducing unwanted bugs. The parts of the logic shared with the VM workflow should be extracted in separate functions and reused.

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if _, err := ctrl.CreateOrUpdate(ctx, r.Client, &pod, func() error {
+		return ctrl.SetControllerReference(labInstance, &pod, r.Scheme)
+	}); err != nil {
+		setLabInstanceStatus(r, ctx, "Could not create GUIContainerPod "+pod.Name+" in namespace "+pod.Namespace, "Warning", "GUIContainerPodNotCreated", labInstance, "", "")
+		return err
+	} else {
+		setLabInstanceStatus(r, ctx, "GUIContainerPod "+pod.Name+" correctly created in namespace "+pod.Namespace, "Normal", "GUIContainerPodCreated", labInstance, "", "")
+	}
```
giorio94, line 172: This is not the correct usage of the CreateOrUpdate function, as it works in case the resource is created but not if it is updated.

I would suggest to have a look at [this blog post](https://engineering.pivotal.io/post/gp4k-kubebuilder-lessons/), which explains why it is so and what is the correct pattern that should be used. Additionally, you can also refer to the [official documentation](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.0/pkg/controller/controllerutil#CreateOrUpdate) and to other examples in the [tenant controller](https://github.com/netgroup-polito/CrownLabs/blob/master/operators/pkg/tenant-controller/tenant_controller.go).

giorio94 (commented):
@qfce I know this PR isn't ready yet. Just some early comments of the most important aspects that should be addressed from my point of view.

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,206 @@
...
+	vmStart time.Time) error {
+	ctx := context.TODO()
+
+	service, ingress, err := CreateInstanceExpositionEnvironment(ctx, r, instance, name, namespace)
+	if err != nil {
+		return err
+	}
+
+	depl := appsv1.Deployment{
+		ObjectMeta: BuildInstanceObjectMeta(name, namespace, environment, instance.Name),
```
giorio94, line 190: Here you should only configure name and namespace. The labels should be set inside the mutate function

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	if _, err := ctrl.CreateOrUpdate(ctx, r.Client, &depl, func() error {
+		depl.Spec = buildContainerInstanceDeploymentSpec(name, instance, environment, r, 6080)
+		return ctrl.SetControllerReference(instance, &depl, r.Scheme)
+	}); err != nil {
+		r.setInstanceStatus(ctx, "Could not create deployment "+depl.Name+" in namespace "+depl.Namespace+": "+err.Error(), "Error", "GUIContainerDeplNotCreated", instance, "", "")
+		klog.Info(err, "")
```
giorio94, line 198: Why klog.Info on error? You should also log something besides the error itself

operators/pkg/instance-controller/status_inspection.go
```diff
@@ -15,6 +16,43 @@ import (
 	crownlabsv1alpha2 "github.com/netgroup-polito/CrownLabs/operators/api/v1alpha2"
 )
 
+func (r *InstanceReconciler) getContainerStatus(ctx context.Context,
```
giorio94, line 19: Why isn't this function based only on the readiness status of the deployment? 

BTW, the best thing would be avoid the loop/sleep approach and return from the reconcile, watching for changes in the deployments. When the deployment becomes ready, the controller awakes, performs a new reconcile and sets the instance ready.

This change would make no longer work the bootTime metric, but I do not feel it a problem for the moment.

operators/pkg/instance-controller/status_inspection.go
```diff
@@ -15,6 +16,43 @@ import (
...
+		port = "22" // SSH
+	}
+
+	err := waitForConnection(host, port)
+	if err != nil {
+		klog.Error(fmt.Sprintf("Unable to check whether %v:%v is reachable", host, port))
+		klog.Error(err)
+	} else {
+		msg := "ContainerDeployment " + depl.Name + " in namespace " + depl.Namespace + " status update to READY."
+		r.setInstanceStatus(ctx, msg, "Normal", "ContainerReady", instance, ip, url)
```
giorio94, line 49: I suppose this should be VmiReady (although it is awkward), otherwise it is not recognized by the dashboard. 
These names will then need to be completely refactored.

operators/pkg/instance-controller/common_logic.go
```diff
@@ -0,0 +1,70 @@
...
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	crownlabsv1alpha2 "github.com/netgroup-polito/CrownLabs/operators/api/v1alpha2"
+	instance_creation "github.com/netgroup-polito/CrownLabs/operators/pkg/instance-creation"
+)
+
+// BuildInstanceObjectMeta creates ObjectMeta for Instance.
+func BuildInstanceObjectMeta(
```
giorio94, line 17: There is already a function configuring the labels [here](https://github.com/netgroup-polito/CrownLabs/blob/42968262487ce25c11be9d5c62651727f1bb0d17/operators/pkg/instance-creation/creation.go#L156)

operators/pkg/instance-controller/common_logic.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+	}
+}
+
+// CreateInstanceExpositionEnvironment creates necessary instance components for accessing the instance (ingress, service and oauth).
+func CreateInstanceExpositionEnvironment(
+	ctx context.Context,
+	r *InstanceReconciler,
+	instance *crownlabsv1alpha2.Instance,
+	name string, namespace string,
```
giorio94, line 37: InstanceReconciler should be a receiver and not a parameter:

```suggestion
func (r *InstanceReconciler) CreateInstanceExpositionEnvironment(
	ctx context.Context,
	instance *crownlabsv1alpha2.Instance,
	name string, namespace string,
```

Name and namespace are already available inside the instance IIUC

giorio94 (commented):
Some initial comments about the overall logic. I haven't yet reviewed the tests though

giorio94 (commented):
As for the tests, I feel they should way more detailed, since now they check just the creation of the deployment and do not make any assertions about its content or the presence of the other resources. Still, to avoid delaying to much this PR, these improvements could be deferred to a subsequent PR. For the moment, I would just move the container tests to a separate file, to avoid increasing too much the length of the current one.

operators/pkg/instance-controller/container_logic.go
qcfe: It's actually something I initially used to find out why stuff wasn't working. It shouldn't be required anymore

giorio94: Perfect

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,226 @@
...
+		ip := ""
+		url := ""
+		status := "VmiCreated"
+		if depl.Status.ReadyReplicas > 0 {
+			ip = service.Spec.ClusterIP
+			url = ingress.GetAnnotations()["crownlabs.polito.it/probe-url"]
+			status = "VmiReady"
+		}
+		r.setInstanceStatus(ctx, "Container Deployment "+depl.Name+" in namespace "+depl.Namespace+" status update to "+status, "Normal", status, instance, ip, url)
+		return ctrl.SetControllerReference(instance, &depl, r.Scheme)
```
giorio94, line 219: These operations should be performed outside the mutate function

operators/pkg/instance-controller/controller.go
```diff
@@ -143,17 +150,21 @@ func (r *InstanceReconciler) generateEnvironments(template *crownlabsv1alpha2.Te
...
+			if err := r.CreateContainerEnvironment(instance, &template.Spec.EnvironmentList[i], namespace, name, vmstart); err != nil {
+				return ctrl.Result{}, err
+			}
 		}
 	}
 	return ctrl.Result{}, nil
 }
 
 // SetupWithManager registers a new controller for Instance resources.
+// Also Deployments are watched in order to better handle container environment
```
giorio94, line 162: Would avoid this comment here. Probably it would be better placed just above the Owns call.

operators/pkg/instance-controller/controller_test.go
```diff
@@ -290,6 +291,215 @@ var _ = Describe("Instance Operator controller", func() {
 	})
 })
 
+var _ = Describe("Instance Operator controller for containers", func() {
```
giorio94, line 294: Please, move to a separate file since this one is becoming too long

operators/pkg/instance-controller/controller_test.go
```diff
@@ -290,6 +291,215 @@ var _ = Describe("Instance Operator controller", func() {
...
+				Name:      InstanceName,
+				Namespace: InstanceNamespace,
+			}, &depl, BeTrue(), timeout, interval)
+
+			By("Checking the pod list size is not zero")
+			podList := v1.PodList{}
+			Expect(k8sClient.List(ctx, &podList, &client.ListOptions{
+				Namespace: InstanceNamespace,
+			})).Should(Succeed())
+			Expect(podList.Size()).ShouldNot(BeZero())
```
giorio94, line 436: This could be avoided, since here you are testing whether Kubernetes works and not your code

operators/pkg/instance-controller/controller_test.go
```diff
@@ -4,21 +4,22 @@ import (
 	"context"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	errors2 "k8s.io/apimachinery/pkg/api/errors"
```
giorio94, line 9:
```suggestion
	errors "k8s.io/apimachinery/pkg/api/errors"
```

giorio94 (commented):
A couple of more comments, mostly related to minor aspects.

When this is fixed and the test pass, we can test the entire workflow in the cluster

operators/pkg/instance-controller/container_logic.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+}
+
+// CreateContainerEnvironment implements the logic to create all the different
+// Kubernetes resources required to start a containerized CrownLabs environment.
+func (r *InstanceReconciler) CreateContainerEnvironment(
+	instance *crownlabsv1alpha2.Instance,
+	environment *crownlabsv1alpha2.Environment,
+	namespace string,
+	name string,
```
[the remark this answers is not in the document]
qcfe: That's true, I kept them because it was also in the vm environment. Should I change those as well?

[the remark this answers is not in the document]
giorio94: As you prefer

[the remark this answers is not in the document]
qcfe: I just saw that name gets '.' characters replaced with '-' on [lin3 143](https://github.com/qcfe/CrownLabs/blob/45ee8c4aefd64f504ad29199adff4ccbaa61d805/operators/pkg/instance-controller/controller.go#L143). Should that operation be done every time I get the instance name? Is there any better way to handle this?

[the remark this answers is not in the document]
giorio94: Sorry, forgot about that. Then just keep the parameter and there should be no problem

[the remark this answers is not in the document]
qcfe: Still this is somehow weird, I mean, is there any particular reason for which names could contain dots? Isn't it possible to ensure such substitution when the name is set?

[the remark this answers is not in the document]
giorio94: No, this is not possible. Resource names can contain dots but labels cannot: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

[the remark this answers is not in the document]
qcfe: Ok, then I'll revert everything

operators/deploy/instance-operator/templates/_helpers.tpl
```diff
@@ -66,3 +66,10 @@ Metrics selector additional labels
 {{- define "instance-operator.metricsAdditionalLabels" -}}
 app.kubernetes.io/component: metrics
 {{- end }}
+
+{{/*
+The labels to be used for sidecar containers images
```
giorio94, line 71:
```suggestion
The tag to be used for sidecar containers images
```

giorio94 (approved):
/lgtm

palexster (approved):

PR #392, opened 2021-01-02
Add finalizers to tenant operator and other improvements

GabriFila:
# Description

This PR includes:
- fix issue when getting users in keycloak in case the username of a user is part of a username of other users
- add labels to tenant resource for firstName and lastName
- make `ready` field required in API declaration to allow proper visualization in case field is false (useful when using `kubectl get tenants | grep false` for first glance feedback)
- add support for finalizers in tenant controller, to allow proper deletion of internal/external resources
- add tests for cluster resources creation. This should not increase the coverage but should be useful in case something changes in the future.

# How Has This Been Tested?

This PR adds tests to:
- check the correct addition of the finalizer reference to the workspace/tenant
- checks that the pair `mariò ròssì` gets turned into `mari rss` when adding the labels for firstName and lastName to the tenant


giorio94 (approved):
/lgtm

palexster (approved):

PR #393, opened 2021-01-05
Fix wrong VM instance generation

palexster:
# Description

This PR fixes the memory quantity associated to a VM instance in the instance operator

# How Has This Been Tested?

Unit Tests

### DIFF
--- a/operators/pkg/instance-creation/creation.go
+++ b/operators/pkg/instance-creation/creation.go
@@ -212,8 +212,8 @@ func CreateOrUpdate(c client.Client, ctx context.Context, object interface{}) er
 }
 
 func CheckLabels(ns v1.Namespace, matchLabels map[string]string) bool {
-	for key := range matchLabels {
-		if _, ok := ns.Labels[key]; !ok {
+	for key, value := range matchLabels {
+		if v1, ok := ns.Labels[key]; !ok || v1 != value {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


giorio94 (approved):

GabriFila:
This PR seems ok. I just noticed 2 things:
1. the PR is composed by3 commits with the same name (I believe some changes in some commits are related only to linting). 
2.  the `TestCheckLabels` is ok, but, just to check, I didn't find any place in the instance controller/creation code that adds those kinds of labels (`crownlabs.polito.it/operator-selector`). My point is not related to the test itself since it tests the checking of labels not the adding of a label, but maybe you want to actually add it also to the resources created by the operator and it is not present at the moment

palexster:
> This PR seems ok. I just noticed 2 things:
> 
> 1. the PR is composed by3 commits with the same name (I believe some changes in some commits are related only to linting).

Fixed thanks!

> 2. the `TestCheckLabels` is ok, but, just to check, I didn't find any place in the instance controller/creation code that adds those kinds of labels (`crownlabs.polito.it/operator-selector`). My point is not related to the test itself since it tests the checking of labels not the adding of a label, but maybe you want to actually add it also to the resources created by the operator and it is not present at the moment

The label is not added by the controller itself, and the instance controller just checks if the namespace is eligible. However, I think we should define a cross-operator label strategy and implement it consistently across the different operator. 

I would not include this in the PR, but I am opening an issue about that.

GabriFila (approved):
Lgtm

PR #395, opened 2021-01-07
Apply kube-score to the rendered manifests

giorio94:
# Description

This PR updates the kube-score check, to apply it to the rendered manifests instead of the old ones.
The `pull_request_target` check fails, since it is still operating on the old files that have been deleted.


GabriFila (approved):
lgtm

palexster (approved):

PR #399, opened 2021-01-12
Bump the automerge action version

giorio94:
# Description

This PR bumps the automerge action version. 
Specifically, the new version is based on JS instead of docker, and thus it is much faster to execute.


GabriFila (approved):
lgtm

palexster (approved):

PR #403, opened 2021-01-18
Dashboard Update

nappozord:
# Description

This PR updates the dashboard to the latest version. Also fixes a bug where sometimes the tenant would not be registered, causing the dashboard to not look for related images and instances (an so make it look like the two lists are loading indefinitely).


giorio94 (approved):

giorio94:
/hold

The dashboard version deployed in the staging environment has some problems

frisso (approved):

PR #409, opened 2021-01-29
Fix deletion bug in tenant operator

GabriFila:
# Description

This PR includes:
- _fix the deletion logic regarding the finalizers for the tenant operator_. When a resource was deleted and the deletion of external resources (keycloak/nextcloud) failed, the operator removed the finalizers from the resource regardless of the outcome of the previous deletion, making the finalizers useless.
- _fix the interaction between the tenant operator and nextcloud_. The problem was related to the authentication management of nextcloud: after 30 minutes of using the API with HTTP basic auth to authenticate the tenant-operator account to edit nextcloud user, the server required the password confirmation of the user (a bit of a 2nd WTF for nextcloud 😅). By forcing go resty client to not store cookies every request results as new one from nextcloud and gets accepted regardless of the 30 minutes

# How Has This Been Tested?

This has not been tested,.
- The first behavior showed when, after having deleted a tenant, nextcloud had some errors and the tenant was successfully deleted by the garbage collection. The tenats had still the corresponding users in nextcloud.
- The second fix did not result in any other issues when creating/updating users, it can be seen in the logs of the tenant operator in the namespace `tenant-409`

### DIFF
--- a/operators/pkg/tenant-controller/nextcloud_utils.go
+++ b/operators/pkg/tenant-controller/nextcloud_utils.go
@@ -95,6 +95,7 @@ func (ncA *NcActor) UpdateUserData(username, param, value string) error {
 		klog.Errorf("Error when sending request for %s update in nextcloud for user %s -> %s", param, username, err)
 		return err
 	}
+
 	statusCode, message, err := parseOCSResponseMeta(res.Body())
 	if err != nil {
 		klog.Errorf("Error when parsing response of %s update in nextcloud of user %s -> %s", param, username, err)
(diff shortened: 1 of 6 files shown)


operators/pkg/tenant-controller/nextcloud_utils.go
```diff
@@ -32,6 +32,7 @@ var ncHeaders = map[string]string{"OCS-APIRequest": "true"}
 // GetUser gets the user with the corresponding username in nextcloud. It returns info about the existence of the user, the displayname of the user and if there are any errors
 func (ncA *NcActor) GetUser(ncUsername string) (found bool, displayname *string, err error) {
 	userEndpoint := ncA.buildOCSEndpoint(fmt.Sprintf("/users/%s", ncUsername))
+	ncA.Client.SetCookieJar(nil)
```
giorio94, line 35: I feel it may be enough to configure this once, when you initialize the client. WDYT?

giorio94 (approved):
/lgtm

operators/pkg/tenant-controller/nextcloud_utils.go
GabriFila: Yep you are right!

GabriFila: I deployed and it works

palexster (approved):

PR #419, opened 2021-02-12
Add comments to the operators APIs for self-documentation

giorio94:
# Description

This PR comments the different CrownLabs APIs, in order to add the documentations to the CRDs, that can later be accessed through `kubectl explain` and the graphical dashboard. Additionally, it enables mandatory comments for all exported functions, and fixes the problems detected by the linter. Finally, the last commit introduces two new fields in the template and instance APIs. which are required for the persistent VMs workflow. Specifically, the introduced fields are:
* `Disk`: as part of the environment resources of the template, to specify the size of the persistent disk.
* `Running`: as part of the instance spec, to allow stopping and restarting persistent VMs.

Fixes # (issue)

# How Has This Been Tested?

- [X] Automated testing
- [X] Linting
- [X] Checking the outcome of `kubectl explain` using a kind cluster


operators/cmd/tenant-operator/main.go
```diff
@@ -152,7 +152,7 @@ func main() {
 	}
 }
 
-// newKcActor sets up a keycloak client with the specififed parameters and performs the first login
+// newKcActor sets up a keycloak client with the specififed parameters and performs the first login.
```
frisso, line 155:
```suggestion
// newKcActor sets up a keycloak client with the specified parameters and performs the first login.
```

operators/cmd/tenant-operator/main.go
```diff
@@ -175,7 +175,7 @@ func newKcActor(kcURL, kcUser, kcPsw, targetRealmName, targetClient, loginRealm
 	}, nil
 }
 
-// checkAndRenewTokenPeriodically checks every intervalCheck if the token is about in less than expireLimit or is already expired, if so it renews it
+// checkAndRenewTokenPeriodically checks every intervalCheck if the token is about in less than expireLimit or is already expired, if so it renews it.
```
frisso, line 178: is about in less ==> ?

operators/samples/tenant.yml
```diff
@@ -1,7 +1,7 @@
 apiVersion: crownlabs.polito.it/v1alpha1
 kind: Tenant
 metadata:
-  name: mariorossi
+  name: mario.rossi
```
frisso, line 4:
```suggestion
  name: john.doe
```
(better to use English terms)

operators/api/v1alpha1/tenant_types.go
```diff
@@ -23,65 +23,89 @@ import (
...
-	// list of the public keys of the tenant for SSH access
+	// The list of the SSH public keys associated with the Tenant. These will be
+	// used to enable to access the remote environments through the SSH protocol.
 	PublicKeys []string `json:"publicKeys,omitempty"`
 
-	// should the resource create the sandbox namespace for k8s practice environment
+	// Whether a sandbox namespace should be created to allow the Tenant play
+	// with Kubernetes.
 	// +kubebuilder:default=false
 	CreateSandbox bool `json:"createSandbox,omitempty"`
```
GabriFila, line 77: _Sorry for the long comment_

_I mark this line, but there are more lines with the following issue:_

Placing the doc comment attached to the `//kubebuilder` marker, adds the marker inside the docs (you can see it if you hover over the filed on vscode). IMHO it is better not to have the marker in the docs. 

**current**
wrong

**fixed**
right

The latter is produced with the following code:
```go
// +kubebuilder:default=false

// Whether a sandbox namespace should be created to allow the Tenant play
// with Kubernetes.
CreateSandbox bool `json:"createSandbox,omitempty"`
```

This code doesn't cancel the effect of the marker (it can be checked by search the word default in the generated CRD, it is still there)

operators/pkg/bastion-controller/bastion_controller.go
```diff
@@ -31,16 +32,16 @@ import (
...
+// BastionReconciler reconciles a Bastion object.
 type BastionReconciler struct {
 	client.Client
 	Log                logr.Logger
 	Scheme             *runtime.Scheme
 	AuthorizedKeysPath string
 }
 
+// Reconcile reconciles the SSH keys of a Tenant resource.
 // +kubebuilder:rbac:groups=crownlabs.polito.it,resources=tenants,verbs=get;watch
```
GabriFila, line 44: I think we don't use these markers in our deploy approach, IMHO this can be the right PR to remove them from all the controllers

operators/api/v1alpha1/tenant_types.go
giorio94: Thanks @GabriFila, did not know about this issue. Should be fixed now

operators/pkg/bastion-controller/bastion_controller.go
giorio94: Definitely.

palexster (approved):

GabriFila (approved):
lgtm

PR #420, opened 2021-02-13
Policies: fix the instance-template-reference policy and improve testing

giorio94:
# Description

This PR improves the *instance-template-reference* policy to fix the problem mentioned in #415. Additionally, it introduces new tests to prevent regressions and ensure those cases are correctly handled (and slightly modifies the others to improve clarity, at least IMO).

Secifically, the fix involves two main problems:
* The policy did not return an error in certain conditions if the referenced namespace had no templates.
* The policy incorrectly returned an error if the referenced namespace contained multiple templates.

@ChiaraOggeri @sofymunari Could you have a look whether the modifications seem ok to you (seems I cannot specify you as reviewers until you accept the invitation)? I am definitely not a rego/gatekeeper expert.

# How Has This Been Tested?

- [X] Unit testing (adding new tests regarding the identified problematic situations)
- [X] Manual testing with a kind cluster, replicating the templates and instances currently present in CrownLabs.

### DIFF
--- a/policies/policies/instance-template-reference.rego
+++ b/policies/policies/instance-template-reference.rego
@@ -1,30 +1,28 @@
 package crownlabs_instance_template_reference
 
-t_ns := input.review.object.spec["template.crownlabs.polito.it/TemplateRef"].namespace
+tmpl_api_version := "crownlabs.polito.it/v1alpha2"
 
-ns = t_ns {
-	count([t_ns]) > 0
-}
+tmpl_ref_field := "template.crownlabs.polito.it/TemplateRef"
 
-else = "default" {
+# Return the referenced template namespace, in case the field is defined, otherwise "default"
+get_tmpl_namespace(review) = namespace {
+	review.object.spec[tmpl_ref_field].namespace
+	namespace := review.object.spec[tmpl_ref_field].namespace
+} else = "default" {
 	true
 }
 
-violation[{"msg": msg, "details": {}}] {
-	not data.inventory.namespace[ns]
-	msg := sprintf("Namespace %v does not exist", [ns])
+# Return the referenced template name
+get_tmpl_name(review) = name {
+	name := review.object.spec[tmpl_ref_field].name
 }
 
-violation[{"msg": msg, "details": {}}] {
-	var := data.inventory.namespace[ns]["crownlabs.polito.it/v1alpha2"].Template
-	var == {}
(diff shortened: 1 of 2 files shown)


policies/verify-instances-templates-reference/policies/rego_rule.rego
```diff
@@ -1,33 +1,32 @@
...
-	
-}
+tmpl_ref_field := "template.crownlabs.polito.it/TemplateRef"
 
-else = "default" {
-	true
+# Return the referenced template namespace, in case the field is defined
+get_tmpl_namespace(review) = namespace {
+	review.object.spec[tmpl_ref_field].namespace
+	namespace := input.review.object.spec[tmpl_ref_field].namespace
```
sofymunari, line 10: Since the function receives "review" as input parameter, it could be directly used instead of "input.review". Anyways, the result should be the same.

giorio94: Fixed, thanks!

policies/verify-instances-templates-reference/policies/rego_rule.rego
```diff
@@ -1,33 +1,32 @@
...
+	namespace := review.object.spec[tmpl_ref_field].namespace
 }
 
-violation[{"msg": msg, "details": {}}] {
-	not data.inventory.namespace[ns]
-	msg := sprintf("Namespace %v doesn't exist", [data.inventory])
+# Return default, in case the namespace field is not defined
+get_tmpl_namespace(review) = namespace {
+	not review.object.spec[tmpl_ref_field].namespace
+	namespace := "default"
```
sofymunari, line 16: To avoid replicating code, maybe it could be a good idea to use a if-else construct instead of using the same function twice. 
```suggestion
	get_tmpl_namespace(review) = namespace {
	review.object.spec[tmpl_ref_field].namespace
	namespace := input.review.object.spec[tmpl_ref_field].namespace 
    }else = "default"{
	true
   }
```

giorio94: Thanks, I modified the rule according to your suggestion. I honestly feel both solutions are pretty equivalent (also in terms of LOC), but it definitely makes sense to use the if-else construct.

frisso (approved):

palexster (approved):

PR #436, opened 2021-04-11
Add support for file manager sidecar container (File Browser)

lucafrancescato:
# Description

This PR includes the support to a file manager sidecar container to run along with the containers of the container environment instance. In particular, the file manager is implemented using [File Browser](https://github.com/filebrowser/filebrowser).
The idea is to provide a web-based interface that lets the user:
- upload files to be consumed by the main application container, and
- download files produced by the main application container.

# How Has This Been Tested?

Ran as a deployment on the cluster.

### DIFF
--- a/operators/cmd/instance-operator/main.go
+++ b/operators/cmd/instance-operator/main.go
@@ -68,6 +68,8 @@ func main() {
 	var containerEnvVncImg string
 	var containerEnvWebsockifyImg string
 	var containerEnvNovncImg string
+	var containerEnvFileBrowserImg string
+	var containerEnvFileBrowserImgTag string
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
@@ -85,6 +87,8 @@ func main() {
 	flag.StringVar(&containerEnvVncImg, "container-env-vnc-img", "crownlabs/tigervnc", "The image name for the vnc image (sidecar for graphical container environment)")
 	flag.StringVar(&containerEnvWebsockifyImg, "container-env-websockify-img", "crownlabs/websockify", "The image name for the websockify image (sidecar for graphical container environment)")
 	flag.StringVar(&containerEnvNovncImg, "container-env-novnc-img", "crownlabs/novnc", "The image name for the novnc image (sidecar for graphical container environment)")
+	flag.StringVar(&containerEnvFileBrowserImg, "container-env-filebrowser-img", "filebrowser/filebrowser", "The image name for the filebrowser image (sidecar for gui-based file manager)")
+	flag.StringVar(&containerEnvFileBrowserImgTag, "container-env-filebrowser-img-tag", "latest", "The tag for the FileBrowser container (the gui-based file manager)")
 	klog.InitFlags(nil)
 	flag.Parse()
 
@@ -114,10 +118,12 @@ func main() {
 		OidcClientSecret:   oidcClientSecret,
 		OidcProviderURL:    oidcProviderURL,
 		ContainerEnvOpts: instance_controller.ContainerEnvOpts{
-			ImagesTag:     containerEnvSidecarsTag,
-			VncImg:        containerEnvVncImg,
-			WebsockifyImg: containerEnvWebsockifyImg,
-			NovncImg:      containerEnvNovncImg,
+			ImagesTag:         containerEnvSidecarsTag,
+			VncImg:            containerEnvVncImg,
+			WebsockifyImg:     containerEnvWebsockifyImg,
+			NovncImg:          containerEnvNovncImg,
+			FileBrowserImg:    containerEnvFileBrowserImg,
+			FileBrowserImgTag: containerEnvFileBrowserImgTag,
 		},
 	}).SetupWithManager(mgr); err != nil {
 		klog.Fatal(err, "unable to create controller", "controller", "Instance")
--- a/operators/pkg/instance-controller/common_logic.go
+++ b/operators/pkg/instance-controller/common_logic.go
@@ -13,22 +13,37 @@ import (
 )
 
 // CreateInstanceExpositionEnvironment creates the components necessary to access the environment (service, ingress and oauth2-proxy related resources).
+// Additionally, it makes the service expose another port and creates an ingress for FileBrowser sidecar container (only for container environments).
 func (r *InstanceReconciler) CreateInstanceExpositionEnvironment(
 	ctx context.Context,
 	instance *crownlabsv1alpha2.Instance,
-	name string,
-) (v1.Service, networkingv1.Ingress, error) {
+	name string, hasFileBrowser bool,
+) (v1.Service, networkingv1.Ingress, string, error) {
 	// create Service to expose the pod
 	service := instance_creation.ForgeService(name, instance.Namespace)
+
+	fileBrowserPortName := "filebrowser"
+	if hasFileBrowser {
+		service.Spec.Ports = append(service.Spec.Ports, v1.ServicePort{
+			Name:     fileBrowserPortName,
+			Protocol: v1.ProtocolTCP,
+			Port:     8080,
+		})
+	}
 
 	if _, err := ctrl.CreateOrUpdate(ctx, r.Client, &service, func() error {
 		return ctrl.SetControllerReference(instance, &service, r.Scheme)
 	}); err != nil {
 		r.setInstanceStatus(ctx, "Could not create service "+service.Name+" in namespace "+service.Namespace+": "+err.Error(), "Error", "ServiceNotCreated", instance, "", "")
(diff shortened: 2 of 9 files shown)


operators/cmd/instance-operator/main.go
```diff
@@ -85,6 +86,7 @@ func main() {
 	flag.StringVar(&containerEnvVncImg, "container-env-vnc-img", "crownlabs/tigervnc", "The image name for the vnc image (sidecar for graphical container environment)")
 	flag.StringVar(&containerEnvWebsockifyImg, "container-env-websockify-img", "crownlabs/websockify", "The image name for the websockify image (sidecar for graphical container environment)")
 	flag.StringVar(&containerEnvNovncImg, "container-env-novnc-img", "crownlabs/novnc", "The image name for the novnc image (sidecar for graphical container environment)")
+	flag.StringVar(&containerEnvFileMgrImg, "container-env-filemanager-img", "filebrowser/filebrowser", "The image name for the filebrowser image (sidecar for graphical container environment)")
```
frisso, line 89:
```suggestion
	flag.StringVar(&containerEnvFileMgrImg, "container-env-filemanager-img", "filebrowser/filebrowser", "The image name for the filebrowser image (sidecar for gui-based file manager)")
```

frisso (approved):

operators/pkg/instance-controller/common_logic.go
```diff
@@ -13,19 +13,28 @@ import (
 )
 
 // CreateInstanceExpositionEnvironment creates the components necessary to access the environment (service, ingress and oauth2-proxy related resources).
+// Additionally, it makes the service expose another port and creates an ingress for file manager sidecar container (only for container environments).
 func (r *InstanceReconciler) CreateInstanceExpositionEnvironment(
 	ctx context.Context,
 	instance *crownlabsv1alpha2.Instance,
-	name string,
-) (v1.Service, networkingv1.Ingress, error) {
+	name string, hasFileManager bool,
```
giorio94, line 20:
```suggestion
	name string, hasFileBrowser bool,
```

I would name it *file browser* instead of *file manager*, for the sake of clarity. The same for all the other references.

operators/pkg/instance-controller/container_logic.go
```diff
@@ -146,6 +147,26 @@ func buildContainerInstanceDeploymentSpec(
 			// LivenessProbe:   &tigerVncProbe,
 			ReadinessProbe: &tigerVncProbe,
 		},
+		{
+			Name:            "filebrowser",
+			Image:           o.FileMgrImg + ":v2.15.0",
```
giorio94, line 152: The tag should not be hard-coded but obtained as a parameter

operators/pkg/instance-controller/container_logic.go
```diff
@@ -146,6 +147,26 @@ func buildContainerInstanceDeploymentSpec(
...
+			Resources:       buildResRequirements(1, 50, resource.MustParse("2G")),
```
giorio94, line 153: Are these resources correct? Those seem extremely high to me.

operators/pkg/instance-controller/container_logic.go
```diff
@@ -192,11 +234,45 @@ func (r *InstanceReconciler) CreateContainerEnvironment(
...
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cmName,
+			Namespace: namespace,
+		},
+		Data:      cmData,
+		Immutable: pointer.Bool(true),
+	}
```
giorio94, line 267: This should be extracted to a separate function to simplify future extensions.

operators/pkg/instance-controller/container_logic.go
```diff
@@ -192,11 +234,45 @@ func (r *InstanceReconciler) CreateContainerEnvironment(
...
 	}
 
+	cmName := "filebrowser-config"
+
+	fileManagerPort := int32(8080)
+	mountPath := "/mydrive"
+
+	cmData := make(map[string]string)
+	cmData["port"] = fmt.Sprintf("%d", fileManagerPort)
+	cmData["baseURL"] = "/" + urlUUID + "/files"
```
giorio94, line 249:
```suggestion
	cmData["baseURL"] = "/" + urlUUID + "/mydrive"
```

IIRC

operators/pkg/instance-controller/container_logic.go
```diff
@@ -192,11 +234,45 @@ func (r *InstanceReconciler) CreateContainerEnvironment(
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cmName,
+			Namespace: namespace,
+		},
+		Data:      cmData,
+		Immutable: pointer.Bool(true),
+	}
+
+	if _, err := ctrl.CreateOrUpdate(ctx, r.Client, &cm, func() error {
```
giorio94, line 269: Since you are configuring the ConfigMap as immutable, it becomes useless to leverage the CreateOrUpdate approach. I would just stick to the classical Create function, discarding the error in case the object already exists.

operators/cmd/instance-operator/main.go
giorio94, line 89: As you are introducing new parameters, those should also be added to the instance-operator helm chart, in order to allow for their configuration.

Some pointers:
* Helm global values: https://github.com/netgroup-polito/CrownLabs/blob/master/deploy/crownlabs/values.yaml
* Helm values: https://github.com/netgroup-polito/CrownLabs/blob/master/operators/deploy/instance-operator/values.yaml
* Helm deploy template: https://github.com/netgroup-polito/CrownLabs/blob/master/operators/deploy/instance-operator/templates/deployment.yaml

giorio94 (commented):
Looks mostly ok to me. Just some minor comments inline

operators/pkg/instance-controller/container_logic.go
s275128: Indeed. I have verified the resource consumption on Grafana and I will change them accordingly.

operators/pkg/instance-controller/container_logic.go
s275128: Could the tag be a parameter of the enclosing function, or is it better to define it as a field of the "ContainerEnvOpts" struct?

giorio94: I believe it should a field of the ContainerEnvOpts struct, as it needs to be configured with a command-line parameter through the helm manifest in order to allow for fast upgrades without having to modify the source code

operators/cmd/instance-operator/main.go
s275128: Is it ok to have v2.15.0 as the default value in the Helm values file?

giorio94: I would leave the default to latest, and override it in the actual inventory in CrownOps. Otherwise this will get soon stale

giorio94 (approved):
/lgtm

Just a couple of minor comments inline
Please, squash also all the commits before merging

operators/pkg/instance-controller/container_logic.go
```diff
@@ -146,6 +147,28 @@ func buildContainerInstanceDeploymentSpec(
...
+			Args: []string{
+				"--port=" + fmt.Sprintf("%d", fileBrowserPort),
+				"--root=" + mountPath,
+				"--baseurl=/" + urlUUID + "/mydrive",
+				"--database=/tmp/database.db",
+				"--noauth=true",
+			},
+			SecurityContext: &contSecCtx,
+			Ports: []v1.ContainerPort{{
+				ContainerPort: fileBrowserPort,
```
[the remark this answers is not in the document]
lucafrancescato: To expand on this, I would set the FileBrowser's TargetPort field of service.Spec.Ports to the same name — at the moment, it is omitted to match the Port field.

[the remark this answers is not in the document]
lucafrancescato: Well, that would mean to have yet another variable in common to pass between common_logic.go and container_logic.go files, which may prove to be unwieldy.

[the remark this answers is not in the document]
lucafrancescato: Please let me know if I just should stick with your proposed solution.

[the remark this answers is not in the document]
giorio94: Not sure if I understood correctly your comment, but currently the port is hard coded, so switching to names would make no difference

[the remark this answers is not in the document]
lucafrancescato: To explain myself better, I will quote the docs about service TargetPort field: "[...] If this is a string, it will be looked up as a named port in the target Pod's container ports.".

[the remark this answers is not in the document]
giorio94: Ok, I think I got your message. Generally speaking, named ports in services are slightly better, as more meaningful than just numbers, However, it is not a big deal in this case, you can stick to the actual approach.

PR #442, opened 2021-04-22
Tune FileBrowser resources and add annotations to FileBrowser ingress

lucafrancescato:
# Description

This PR aims to fine-tune the resources (requests and limits) for the FileBrowser container.

Moreover, it adds some useful annotations to the dedicated FileBrowser ingress of the container-based Instance. These annotations remove the limit of 1 MB for the maximum allowed size of the client request body, as well as increase connection timeouts between two requests or responses, and other minor tweaks that overall increase the final user experience when dealing with FileBrowser.


giorio94 (approved):
/lgtm

palexster (approved):

PR #448, opened 2021-04-26
Fix blender 2.92 compatibility issue with Ubuntu

qcfe:
# Description

This PR provides a single Dockerfile for Blender container images and uses build args to specify the Blender version to ship into the image. 
This also fixes an issue with the Blender 2.92 image, which has been unnoticed up to now. For some unknown reason Ubuntu 18.04 has some compatibility issues with Blender > 2.83.8. The unified Dockerfile builds the image from scratch using Ubuntu 20.04.

# How Has This Been Tested?
The imagea has been run in a compatible local Docker environment.


giorio94 (commented):
It looks to me this Dockerfile is identical to the LTS one, with just a different version. If so, wouldn't it be possible to keep only one and build twice with a different version as argument? Just to prevent code duplication and unwanted misalignements

qcfe:
@giorio94 it is identical indeed but I didn't look close enough in the `build.yaml` file to see it could be authomatized easily, I'll proceed to make the change

giorio94:
You could use build-args as here: https://github.com/netgroup-polito/CrownLabs/blob/df41684c2730ed583a82053a4ae61ab95b2f1816/.github/workflows/build.yml#L122

giorio94 (approved):
/lgtm

GabriFila (approved):

PR #452, opened 2021-04-30
Add qlkube Helm charts

qcfe:
# Description

This PR adds the generated helm charts for the deployment of qlkube.

### DIFF
--- a/qlkube/deploy/qlkube/Chart.yaml
+++ b/qlkube/deploy/qlkube/Chart.yaml
@@ -1,0 +1,20 @@
+apiVersion: v2
+name: qlkube
+description: The component exposing the Kubernetes resources through graphql
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+icon: https://crownlabs.polito.it/images/logo.svg
--- a/qlkube/deploy/qlkube/values.yaml
+++ b/qlkube/deploy/qlkube/values.yaml
@@ -1,0 +1,48 @@
+# Default values for qlkube.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: crownlabs/qlkube
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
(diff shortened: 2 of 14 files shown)


qlkube/deploy/qlkube/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v2
+name: qlkube
+description: A Helm chart for Kubernetes
```
giorio94, line 3:
```suggestion
description: The component exposing the Kubernetes resources through graphql
```

Or something like that

qlkube/deploy/qlkube/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+podAnnotations: {}
+
+securityContext:
+  capabilities:
+    drop:
+    - ALL
+  readOnlyRootFilesystem: false
+  runAsNonRoot: true
+  runAsUser: 101
+  runAsGroup: 101
```
giorio94, line 27:
```suggestion
  runAsUser: 1000
  runAsGroup: 1000
```

Since this is the uid of node

qlkube/deploy/qlkube/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+  path: /
+  secret: qlkube-certificate
+
+resources:
+  limits:
+    cpu: 1000m
+    memory: 250Mi
+  requests:
+    cpu: 100m
+    memory: 100Mi
```
giorio94, line 47:
```suggestion
    cpu: 1000m
    memory: 500Mi
  requests:
    cpu: 100m
    memory: 250Mi
```

Looking at the stats of the one deployed in the cluster, I would increase a bit the memory

qlkube/deploy/qlkube/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+  runAsGroup: 101
+  privileged: false
+
+service:
+  type: ClusterIP
+  port: 80
+
+ingress:
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt-production
```
giorio94, line 36: Would also add `nginx.ingress.kubernetes.io/custom-http-errors: "418"` to avoid triggering the custom error page in case of errors.

giorio94 (commented):
Thanks @qcfe, seems mostly ok to me. 

In addition to a couple of minor comments inline, you should also create a service account, which is necessary to qlkube to scrape the APIs at the beginning. Though no rolebindings is necessary, as authentication is sufficient.

Once the PR in CrownOps to fill the inventory is also ready and merged, we can try the deploy-staging and ensure everything works correctly.

qcfe:
Hi @giorio94, thanks for your review. I added the serviceAccount (I think). Is it correct? 
I'm going to start working on the crownOps PR asap

giorio94:
> Hi @giorio94, thanks for your review. I added the serviceAccount (I think). Is it correct?
> I'm going to start working on the crownOps PR asap

The serviceAccount seems ok to me. You just have to reference it to the deployment

qcfe:
I guess I also have to set rolebinding and role somewhere? I'm getting documented about that

giorio94:
> I guess I also have to set rolebinding and role somewhere? I'm getting documented about that

No, in this case you do not need to create them, although typically you would (getting the openAPI schema requires only to be authenticated, and does not need specific authorizations)

GabriFila (approved):
lgtm

giorio94 (approved):
/lgtm

PR #454, opened 2021-05-01
Add decorateBaseSchema function qlkube

Ralls0:
# Description

New function has been added. `decorateBaseSchema` function allow to change the dynamic schema create on the fly by qlkube. This new function uses five parameters:
* @param {*} targetQuery : Query wrapped int the extended object Type
* @param {*} extendedType : Object type that should be extended
* @param {*} argsNeeded : Arguments extracted from the father and passed from the wrapper
* @param {*} nameWrapper :  Custom name for the wrapper
* @param {*} baseSchema : GraphQL Schema where the type and query are written

then a new `GraphQLSchema` object is returned.

# How Has This Been Tested?

Below a test query used to test the new feature:

```graphql
query test {
  itPolitoCrownlabsV1alpha2Instance (
    name: "green-tea-6831",
    namespace: "tenant-john-doe"
  )
  {
    spec {
      running
      templateCrownlabsPolitoItTemplateRef {
        name
        namespace
        templateWrapper{
          itPolitoCrownlabsV1alpha2Template {
            kind
            spec {
              prettyName
            }
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/qlkube/src/decorateBaseSchema.js
+++ b/qlkube/src/decorateBaseSchema.js
@@ -1,0 +1,74 @@
+const { gql } = require('apollo-server-core');
+const { extendSchema } = require('graphql/utilities');
+const { addResolversToSchema } = require('@graphql-tools/schema');
+
+/**
+ *
+ * @param {*} targetQuery : Query wrapped int the extended object Type
+ * @param {*} extendedType : Object type that should be extended
+ * @param {*} baseSchema : GraphQL Schema where the type and query are written
+ * @param {*} nameWrapper : Custom name for the wrapper
+ * @param {*} argsNeeded : Arguments extracted from the father and passed from the wrapper
+ * @returns
+ */
+
+function capitalizeType(name) {
+  return name[0].toUpperCase() + name.slice(1);
+}
+
+module.exports = {
+  decorateBaseSchema: function (
+    targetQuery,
+    extendedType,
+    baseSchema,
+    nameWrapper,
+    argsNeeded
+  ) {
+    if (!targetQuery) throw 'Parameter targetQuery cannot be empty!';
+    if (!extendedType) throw 'Parameter extendedType cannot be empty!';
+    if (!nameWrapper) throw 'Parameter nameWrapper cannot be empty!';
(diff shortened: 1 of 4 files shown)


qlkube/src/decorateBaseSchema.js
```diff
@@ -0,0 +1,65 @@
...
+    argsNeeded,
+    nameWrapper,
+    baseSchema
+  ) {
+    if (!targetQuery) return baseSchema;
+    if (!extendedType) return baseSchema;
+    const targetType = baseSchema.getQueryType().getFields()[targetQuery];
+    if (!targetType) return baseSchema;
+
+    nameWrapper = nameWrapper ? nameWrapper : 'fieldWrapper';
```
GabriFila, line 32: I would provide an error in case the `nameWrapper` is not defined, as you did with `targetQuery`

qlkube/src/decorateBaseSchema.js
```diff
@@ -0,0 +1,65 @@
...
+            }
+        `;
+    const resolvers = {
+      [extendedType]: {
+        [nameWrapper]: (parent, args, context, info) => {
+          let newParent = {};
+          for (e of argsNeeded) {
+            newParent[e] = parent[e];
+          }
+          return newParent !== {} ? newParent : parent; // gestire errori
```
GabriFila, line 50: The `!== {}` check will always result false, if you need this check you should use a loop with Obejct.keys. But I believe that a better approach would be to throw an error if the needed prop doesn't exist on the parent.

(also please remove the `// gestire errori` part)

GabriFila:
@Ralls0 thanks for the PR! 😍 
I added few comments, also please edit the title of the PR and remove the `testA,testB` part from the PR description

qlkube/src/decorateBaseSchema.js
Ralls0: Being that argsNeeded is optional also newParent could be empty, isn't so?
May be `Object.keys(newParent).length !==0 ? newParent : parent` a good alternative? 
I could add also a default value for the optional agrs.

GabriFila (approved):
lgtm

qcfe (approved):
/lgtm

PR #457, opened 2021-05-05
First components of activePage

Nepec:
# Description
Adding the following components to the activePage section:
- **ActiveTemplates** - accordion in both a simple and nested version. The simple is thought for the student view and contains simple VM instances, while the nested version is thought for the professor view and is divided into courses->termplates->VMs
- **TemplateHeader** - title of each accordion displaying the number of active VM/the number of templates and a "delete all" button. The button has a responsive behavior as it is substituted by a "more" icon when the page shrinks below the `md` breakpoint
- **ActiveLanding** - a generic temporary wrapper to better display the accordion on the local barebones website. It should be placed in the `<Route path="/active"></Route>` comopnent in `App.tsx`

# Notes
The data structures contained in `./activePage/utils/data` are temporary and are used only to showcase the components.

As pointed out in the Slack channel there are issues in the correct rendering of components on Storybook: some Tailwind properties conflict with some Ant properties. I've managed to work around most of them, however there are still issues with:
- background colors not being displayed
- responsive behavior not working: in the current iteration both the "Delete All" button and the "more" icon will be visible at all times

### DIFF
--- a/frontend/src/components/activePage/ActiveLanding/ActiveLanding.stories.tsx
+++ b/frontend/src/components/activePage/ActiveLanding/ActiveLanding.stories.tsx
@@ -1,0 +1,73 @@
+import { Story } from '@storybook/react';
+import { Title, Description, Stories } from '@storybook/addon-docs/blocks';
+import ActiveLanding, { IActiveLandingProps } from './ActiveLanding';
+import { someKeysOf } from '../../../utils';
+
+export default {
+  title: 'Components/ActivePage/ActiveLanding/ActiveLanding',
+  component: ActiveLanding,
+  parameters: {
+    docs: {
+      page: () => {
+        <>
+          <Title />
+          <Description />
+        </>;
+      },
+    },
+  },
+  decorators: [
+    (Story: any) => (
+      <div style={{ height: '500px' }}>
+        <Story />
+      </div>
+    ),
+  ],
+};
+
+const defaultArgs: someKeysOf<IActiveLandingProps> = {
+  isTenantManager: false,
+};
+
+const Template: Story<IActiveLandingProps> = args => (
+  <ActiveLanding {...args} />
+);
+
+export const Default = Template.bind({});
+
+Default.args = defaultArgs;
+Default.parameters = {
+  docs: {
+    // eslint-disable-next-line react/no-multi-comp
+    page: () => {
+      return (
+        <>
+          <Title>Active landing page</Title>
+          <Description>
+            This component showcases the dynamic toggle between **user view**
+            and **manager view**. The switch is going to be visible only for
+            tenants with manager privileges.
+          </Description>
+          <Stories includePrimary />
+        </>
+      );
+    },
+  },
+};
+
+export const Manager = Template.bind({});
+
+Manager.args = { ...defaultArgs, isTenantManager: true };
+Manager.parameters = {
+  docs: {
+    // eslint-disable-next-line react/no-multi-comp
+    page: () => {
+      return (
+        <>
+          <Title>Active landing page - manager</Title>
+          <Stories />
+        </>
+      );
+    },
+  },
+};
--- a/frontend/src/components/activePage/ActiveLanding/ActiveLanding.tsx
+++ b/frontend/src/components/activePage/ActiveLanding/ActiveLanding.tsx
@@ -1,0 +1,39 @@
+import { FC, useState } from 'react';
+import { Col } from 'antd';
+import NestedTables from '../NestedTables/NestedTables';
+import ViewModeButton from '../ActiveLanding/ViewModeButton/ViewModeButton';
+import { workspaces, templates } from '../tempData';
+import Box from '../../common/Box';
+
+export interface IActiveLandingProps {
+  isTenantManager: boolean;
+}
+
+const ActiveLanding: FC<IActiveLandingProps> = ({ ...props }) => {
+  const [managerView, setManagerView] = useState(false);
+  const { isTenantManager } = props;
+  return (
+    <Box
+      header={{
+        size: 'small',
+        right: isTenantManager && (
+          <ViewModeButton
+            setManagerView={setManagerView}
+            managerView={managerView}
+          />
+        ),
+      }}
+    >
+      <Col>
+        <NestedTables
+          workspaces={workspaces}
+          templates={templates}
+          isManager={managerView}
+          destroyAll={() => alert('VMs deleted')}
+        />
+      </Col>
+    </Box>
+  );
+};
+
+export default ActiveLanding;
--- a/frontend/src/components/activePage/ActiveLanding/ViewModeButton/ViewModeButton.tsx
+++ b/frontend/src/components/activePage/ActiveLanding/ViewModeButton/ViewModeButton.tsx
@@ -1,0 +1,23 @@
+import { FC, Dispatch, SetStateAction } from 'react';
+import { Radio } from 'antd';
+
+export interface IViewModeButtonProps {
+  managerView: boolean;
+  setManagerView: Dispatch<SetStateAction<boolean>>;
+}
+
+const ViewModeButton: FC<IViewModeButtonProps> = ({ ...props }) => {
+  const { setManagerView, managerView } = props;
+
+  return (
+    <Radio.Group
+      value={managerView}
+      onChange={e => setManagerView(e.target.value)}
+    >
+      <Radio.Button value={false}>Personal Instances</Radio.Button>
+      <Radio.Button value={true}>Managed Instances</Radio.Button>
+    </Radio.Group>
+  );
+};
+
+export default ViewModeButton;
--- a/frontend/src/components/activePage/NestedTables/NestedTables.tsx
+++ b/frontend/src/components/activePage/NestedTables/NestedTables.tsx
@@ -1,0 +1,106 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
+import { FC, useState } from 'react';
+import { Table } from 'antd';
+import { CaretRightOutlined } from '@ant-design/icons';
+import RowHeading from '../RowHeading/RowHeading';
+
+export interface IWorkspace {
+  key: string;
+  id: string;
+  name: string;
+  templates: Array<ITemplate>;
+}
+
+export interface ITemplate {
+  key: string;
+  id: string;
+  name: string;
+  workspace: string;
+  nActiveInstances: number;
+}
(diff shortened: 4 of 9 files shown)


frontend/src/components/activePage/utils/data.tsx
```diff
@@ -0,0 +1,53 @@
+export interface IVM {
```
GabriFila, line 1: Since you use this file inside a single component, decelare the needed interface there and if needed, import them in the stories

frontend/src/components/activePage/ActiveTemplates/ActiveTemplates.tsx
```diff
@@ -0,0 +1,94 @@
...
+  nested: boolean;
+}
+
+const ActiveTemplates: FC<IActiveTemplatesProps> = ({ ...props }) => {
+  const { templates, courses, destroyAll, nested } = props;
+  const { Panel } = Collapse;
+  return (
+    <Collapse
+      accordion
+      className="bg-gray-200 border-0 mx-10"
```
GabriFila, line 20: If you look at the storybook, when in mobile view, the `mx-10` is too much. I would reduce it in that case

frontend/src/components/activePage/ActiveTemplates/ActiveTemplates.tsx
```diff
@@ -0,0 +1,94 @@
+import { FC } from 'react';
+import { Collapse } from 'antd';
+import { CaretRightOutlined } from '@ant-design/icons';
+import TemplateHeader from '../TemplateHeader/TemplateHeader';
+import { ITemplate, ICourse } from '../utils/data';
+
+export interface IActiveTemplatesProps {
+  courses: Array<ICourse>;
```
GabriFila, line 8: Change courses to workspaces

frontend/src/components/activePage/ActiveTemplates/ActiveTemplates.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Title, Description, Stories } from '@storybook/addon-docs/blocks';
+import { someKeysOf } from '../../../utils';
+import { courses, templates } from '../utils/data';
+
+export default {
+  title: 'Components/ActivePage/ActiveTemplates',
+  component: ActiveTemplates,
+  argTypes: { destroyAll: { action: 'clicked' } },
+  parameters: {
+    docs: {
```
GabriFila, line 13: It is great you added the docs to the storry!

If you want and when you have time, you could make another PR where you add instructions in the guidelines on how to do it. The docs on the stories would still be optional, but if someone would like to add them, he/she would already know how to do it

GabriFila (changes requested):

frontend/src/components/activePage/ActiveLanding/ActiveLanding.tsx
```diff
@@ -0,0 +1,94 @@
+import { FC, useState } from 'react';
+import { Col, Switch } from 'antd';
+import WorkspacesTable from '../Workspaces/WorkspacesTable';
+import TemplatesTable from '../Templates/TemplatesTable';
+import Box from '../../common/Box';
+
+export interface IWorkspace {
```
GabriFila, line 7: This data should be placed in the story. for the moment pass it as a prop. Later it may become a context

frontend/src/components/activePage/ActiveLanding/ActiveLanding.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <>
+          <Title />
+          <Description />
+        </>;
+      },
+    },
+  },
+  decorators: [
+    (Story: any) => (
+      <div style={{ margin: '2rem', height: '500px' }}>
```
GabriFila, line 20: Why do you use this decorator? Also, on mobile/tablet it doesn't fit well

frontend/src/components/activePage/ActiveLanding/ActiveLanding.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const ActiveLanding: FC = () => {
+  const [managerView, setManagerView] = useState(false);
+  return (
+    <Box
+      header={{
+        size: 'small',
+        center: (
+          <div className="flex justify-end ml-auto">
+            <Switch
```
GabriFila, line 69: Could you please look for an icon that explain a bit more the concept of manager/user?
Also I would add here the check to hide the switch if the user is never manager.

frontend/src/components/activePage/RowHeading/RowHeading.tsx
```diff
@@ -0,0 +1,71 @@
...
+          }}
+        >
+          Destory All
+        </Button>
+        <Dropdown
+          overlay={
+            <Menu>
+              <Menu.Item icon={<DeleteOutlined className="text-lg" />} danger>
+                {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+                <a
```
GabriFila, line 47: Why did you use an ancora tag for the `DestroyAll`?

frontend/src/components/activePage/Workspaces/WorkspacesTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+        ),
+        // eslint-disable-next-line react/no-multi-comp
+        expandedRowRender: record => (
+          <TemplatesTable
+            templates={record.templates}
+            destroyAll={destroyAll}
+          />
+        ),
+        expandRowByClick: true,
+        indentSize: 100,
```
GabriFila, line 45: I don't think it is strictly related to this line. But check the storybook on mobile, because the workspaces/templates sections become very small and the `...` menu gets pushed on a new line

frontend/src/components/activePage/Templates/TemplatesTable.tsx
```diff
@@ -0,0 +1,70 @@
...
+    >
+      <Column
+        title="Templates"
+        dataIndex="name"
+        key="key"
+        render={(text, record: ITemplate) => {
+          return (
+            <RowHeading
+              text={record.name}
+              nActive={record.nActiveMachines}
```
GabriFila, line 59: Use `nActiveInstances` instead of `nActiveMachines`

frontend/src/components/activePage/ActiveLanding/ActiveLanding.tsx
Nepec: I'm not sure how to organize my data since I need its interfaces to build the `IActiveLandingProps` interface. Should I declare the `Itemplate` and `IWorkspaces` interfaces here and move the actual data structures in the story (which would then import the interfaces from this file)?

frontend/src/components/activePage/Workspaces/WorkspacesTable.tsx
Nepec: As far as the mobile view is concerned I haven't really focused on that screen size as this frontend is Desktop oriented and the mobile view is just too narrow. I've focused around the `lg` and `md` breakpoints thinking of when a user splits his screen between the Crownlabs site and another window.
Regardless, I want to add ellipses to all the template and workspace titles which should mitigate the problem also on the narrowest of views

Also I should remove the `indentSize` line as it does not work at all.

frontend/src/components/activePage/ActiveLanding/ActiveLanding.stories.tsx
Nepec: Just to center the view in the Docs a bit and have a high-enough window to show the entire table without scrolling. I can remove the margin as it actually looks ok without.

GabriFila: I would remove the margins

frontend/src/components/activePage/ActiveLanding/ActiveLanding.tsx
GabriFila: You can export the interfaces. If they are used in a single component declare them there, instead if they are used in multiple components, it is better to declare them in a separate file (still in you working folder).
Then import the interfaces in the stories where you need them.

The point is that the component should have no hardcoded values (unless those that will be actually hardcoded when used, e.g. a particular color/setting). The data you wrote is dynamic from the perspective of the component, so if you want to use it make it a prop (or a context with the necessary decorators in the stories)

GabriFila (approved):

frontend/src/components/activePage/NestedTables/NestedTables.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const accordion = (expanded: boolean, record: rowType) => {
+    isManager
+      ? expanded
+        ? setExpandedID([(record as IWorkspace).key])
+        : setExpandedID([''])
+      : expanded
+      ? setExpandedID([(record as ITemplate).key])
+      : setExpandedID(['']);
+  };
```
QcFe, line 44:
```suggestion
  const accordion = (expanded: boolean, record: rowType) => {
    const expId = !expanded ? '' 
      : (isManager ? (record as IWorkspace).key : (record as ITemplate).key);
    setExpandedID([expId]);
  };
```
This part seems a bit redundant and not too easy to be read. 
I tried to make it cleaner but I don't know if I succeeded. You're totally free to reject this suggestion (especially if it's not correct 😂).

frontend/src/components/activePage/ActiveLanding/ViewModeButton/ViewModeButton.tsx
```diff
@@ -0,0 +1,25 @@
+import { FC, Dispatch, SetStateAction } from 'react';
+import { Radio } from 'antd';
+
+export interface IViewModeButtonProps {
+  textUsr: string;
+  textMngr: string;
```
QcFe, line 6: Do we have any actual need to have these passed from outside? 
(I guess I'm missing something as my usual 🥴)

QcFe (approved):
A couple of minor things but it's good for me!

frontend/src/components/activePage/NestedTables/NestedTables.tsx
Nepec: Oh yes, much better thanks!

frontend/src/components/activePage/ActiveLanding/ViewModeButton/ViewModeButton.tsx
Nepec: Not really, I just thought it would be neat to be able to change the text through a prop.

PR #466, opened 2021-05-13
Add PVC support for docker instances

QcFe:
# Description
This PR adds support for attaching PVC storages onto container-based environment in order to avoid data loss if the pod associated to the instance dies. The PVC is bound to the instance and is deleted when the instance is destroyed. This behaviour applies when a value is provided for the `Disk` field in the templates resources.

# How Has This Been Tested?
- [x] Added an explicit test to see if the PVC is created.

### DIFF
--- a/operators/pkg/instance-controller/container_controller_test.go
+++ b/operators/pkg/instance-controller/container_controller_test.go
@@ -108,10 +108,53 @@ var _ = Describe("Instance Operator controller for containers", func() {
 			},
 			Status: crownlabsv1alpha2.InstanceStatus{},
 		}
+		environmentWithPVC = crownlabsv1alpha2.Environment{
+			Name:            InstanceName,
+			Image:           "crownlabs/pycharm",
+			EnvironmentType: crownlabsv1alpha2.ClassContainer,
+			Resources: crownlabsv1alpha2.EnvironmentResources{
+				CPU:                   1,
+				ReservedCPUPercentage: 1,
+				Memory:                resource.MustParse("1024M"),
+				Disk:                  resource.MustParse("5G"),
+			},
+		}
+		templateWithPVC = crownlabsv1alpha2.Template{
+			TypeMeta: metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      TemplateName + "with-pvc",
+				Namespace: TemplateNamespace,
+			},
+			Spec: crownlabsv1alpha2.TemplateSpec{
+				WorkspaceRef: crownlabsv1alpha2.GenericRef{},
+				PrettyName:   "App Container Template",
+				Description:  "This is the container template",
+				EnvironmentList: []crownlabsv1alpha2.Environment{
+					environmentWithPVC,
+				},
+				DeleteAfter: "30d",
+			},
+			Status: crownlabsv1alpha2.TemplateStatus{},
+		}
+		instanceWithPVC = crownlabsv1alpha2.Instance{
+			TypeMeta: metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      InstanceName + "with-pvc",
+				Namespace: InstanceNamespace,
+			},
+			Spec: crownlabsv1alpha2.InstanceSpec{
+				Template: crownlabsv1alpha2.GenericRef{
+					Name:      TemplateName + "with-pvc",
+					Namespace: TemplateNamespace,
+				},
+			},
+			Status: crownlabsv1alpha2.InstanceStatus{},
+		}
 		ns   = v1.Namespace{}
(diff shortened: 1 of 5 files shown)


operators/pkg/instance-controller/container_logic.go
```diff
@@ -216,6 +216,19 @@ func buildContainerInstanceDeploymentSpec(
...
+		VolumeSource: v1.VolumeSource{},
+	}
+
+	if environment.Resources.Disk.IsZero() {
+		volume.VolumeSource.EmptyDir = &v1.EmptyDirVolumeSource{}
+	} else {
+		volume.VolumeSource.PersistentVolumeClaim = &v1.PersistentVolumeClaimVolumeSource{
+			ClaimName: name,
+		}
+	}
```
giorio94, line 230: Could be moved to an ad-hoc function, to simplify also unit-testing

operators/pkg/instance-controller/container_logic.go
```diff
@@ -256,6 +278,24 @@ func (r *InstanceReconciler) CreateContainerEnvironment(
...
+		pvc := v1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+			},
+			Spec: buildContainerInstancePVCSpec(environment),
+		}
+
+		if _, err := ctrl.CreateOrUpdate(ctx, r.Client, &pvc, func() error {
+			// PVC's specs are immutable, they have to be set directly
```
giorio94, line 291: Even though they are immutable, the correct way would be set them here, checking if the creationTimeStamp is equal to zero (as in the example).

operators/pkg/instance-controller/container_logic.go
```diff
@@ -256,6 +278,24 @@ func (r *InstanceReconciler) CreateContainerEnvironment(
...
+			return ctrl.SetControllerReference(instance, &pvc, r.Scheme)
+		}); err != nil {
+			r.setInstanceStatus(ctx, "Could not create PVC "+pvc.Name+" in namespace "+pvc.Namespace+": "+err.Error(), "Error", "VmiNotCreated", instance, "", "")
+			return err
+		}
```
giorio94, line 296: Please, add a log line also in case of success, as [here](https://github.com/netgroup-polito/CrownLabs/blob/7c77875eca40d237e32f86b625a00a33c6c1e3a4/operators/pkg/tenant-controller/workspace_controller.go#L276), to simplify troubleshooting

operators/pkg/instance-controller/container_controller_test.go
```diff
@@ -168,6 +211,45 @@ var _ = Describe("Instance Operator controller for containers", func() {
...
+			flag := true
+			expectedOwnerReference := metav1.OwnerReference{
+				Kind:               "Instance",
+				APIVersion:         "crownlabs.polito.it/v1alpha2",
+				UID:                instanceWithPVC.ObjectMeta.UID,
+				Name:               InstanceName + "with-pvc",
+				Controller:         &flag,
+				BlockOwnerDeletion: &flag,
+			}
+			Expect(pvc.ObjectMeta.OwnerReferences).To(ContainElement(expectedOwnerReference))
```
giorio94, line 248: Would also check that the PVS has the expected size

giorio94 (commented):
Overall, this looks mostly ok to me. I just left some minor comments inline.

operators/pkg/instance-controller/container_logic.go
QcFe: I tried but if I set the specs inside the CreateOrUpdate it fails saying they're immutable...

operators/pkg/instance-controller/container_logic.go
```diff
@@ -256,6 +288,27 @@ func (r *InstanceReconciler) CreateContainerEnvironment(
...
+			// PVC's spec is immutable, it has to be set at creation
+			if pvc.ObjectMeta.CreationTimestamp.IsZero() {
+				pvc.Spec = buildContainerInstancePVCSpec(environment)
+			}
+			return ctrl.SetControllerReference(instance, &pvc, r.Scheme)
+		}); err != nil {
+			r.setInstanceStatus(ctx, "Could not create PVC "+pvc.Name+" in namespace "+pvc.Namespace+": "+err.Error(), "Error", "VmiNotCreated", instance, "", "")
+			return err
+		}
+		klog.Infof("Successfully created PVC for instance %s in namespace %s", instance, namespace)
```
giorio94, line 309: Nit. In the log, use the return value of CreateOrUpdate instead of the hard-coded `created` value, as in the example

giorio94 (approved):
/lgtm

Just one minor comment, than squash and from my side this is good to go

GabriFila (approved):

palexster (approved):

PR #468, opened 2021-05-14
Fix issue with docker layers caching

giorio94:
# Description

This PR fixes an issue in the configuration of the docker layers caching in the build-and-publish actions, which caused the usage of outdated caches in certain conditions. Essentially, if an exact match occurs, the resulting cache is not updated, although it may have changed due to a new version of the base image. With this modification, the cache is assigned a name guaranteed to be unique, to ensure it is always updated.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,8 +181,13 @@ jobs:
         uses: actions/cache@v2
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-${{ matrix.component }}-buildx-${{ hashFiles(format('{0}/**', matrix.context), matrix.dockerfile) }}
+          # The cache key is composed of the combination of the component name, the hash of the files in the build context and the hash of the commit.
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -181,8 +181,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-${{ matrix.component }}-buildx-${{ hashFiles(format('{0}/**', matrix.context), matrix.dockerfile) }}
+          key: ${{ runner.os }}-${{ matrix.component }}-buildx-${{ hashFiles(format('{0}/**', matrix.context), matrix.dockerfile) }}-${{ github.sha }}
```
frisso, line 184: I would simply add a comment before this line showing an example of the resulting name

frisso (approved):

GabriFila (approved):

QcFe (approved):

PR #469, opened 2021-05-14
SSHKeysForm + SSHKeysTable - Modal form for key addition

luco5826:
# Description

This PR introduces 2 new components for the UserPanel section of the frontend:
- `SSHKeysTable` - A simple wrapper around the list of ssh keys, shown in the SSH Keys panel
- `SSHKeysForm` - A form where the user can add new SSH keys to its list (this component is wrapped by a `Modal` )

Co-authored-by: [Alessandro Bacci](https://github.com/Ale142)

### DIFF
--- a/frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
+++ b/frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
@@ -1,0 +1,124 @@
+import { Input, Form, Button, Row } from 'antd';
+import { FC, useState } from 'react';
+
+export interface ISSHKeysFormProps {
+  onSaveKey: (newKey: { name: string; key: string }) => void;
+  onCancel: () => void;
+}
+
+const acceptedAlgorithms = [
+  'sk-ecdsa-sha2-nistp256@openssh.com',
+  'ecdsa-sha2-nistp256',
+  'ecdsa-sha2-nistp384',
+  'ecdsa-sha2-nistp521',
+  'sk-ssh-ed25519@openssh.com',
+  'ssh-ed25519',
+  'ssh-dss',
+  'ssh-rsa',
+];
+
+const SSHKeysForm: FC<ISSHKeysFormProps> = props => {
+  const { onSaveKey, onCancel } = props;
+  const [validForm, setValidForm] = useState<boolean>(false);
+  const [form] = Form.useForm();
+
+  const cancelForm = () => {
+    form.resetFields();
+    onCancel();
+  };
+  const validateSSHKey = async (rules: any, key: string) => {
+    if (!key) throw new Error('Key field is mandatory');
+    const result = key.split(/\s+/); // split regardless the amount of spaces
+    if (result.length !== 2 && result.length !== 3)
+      throw new Error('Invalid key format');
+    if (result[1] === '') throw new Error('Empty SSH key');
+    if (!acceptedAlgorithms.includes(result[0]))
+      throw new Error('Unrecognized SSH key format');
+  };
+
+  // Used to disable Save button
+  const handleChange = async (
+    _: any,
+    { name, key }: { name: string; key: string }
+  ) => {
+    try {
+      if (key && name) {
+        await form.validateFields(['name', 'key']);
+        setValidForm(true);
+      } else if (key) await form.validateFields(['key']);
+      else await form.validateFields(['name']);
+    } catch (_) {
+      setValidForm(false);
+    }
+  };
+  const submitForm = ({ name, key }: { name: string; key: string }) => {
+    // the name field is appended to the key
+    const comment = key.split(/\s+/)[2];
+    name = name.trim();
+    if (comment) key = `${key}:${name}`;
+    else key = `${key} ${name}`;
+
+    onSaveKey({ name, key });
+    form.resetFields();
+  };
+
+  return (
+    <Form
+      form={form}
+      labelCol={{ span: 4 }}
+      wrapperCol={{ span: 24 }}
+      onFinish={submitForm}
+      onValuesChange={handleChange}
+    >
+      <Form.Item
+        name="name"
+        label="Name"
+        validateTrigger="onBlur"
+        rules={[
+          {
+            required: true,
+            message: 'Field required',
+          },
+        ]}
+      >
+        <Input />
+      </Form.Item>
+      <Form.Item
+        validateTrigger="onBlur"
+        name="key"
+        label="Public Key"
+        rules={[
+          {
+            required: true,
+            validator: validateSSHKey,
+          },
+        ]}
+      >
+        <Input.TextArea
+          rows={10}
+          placeholder={`Your PUBLIC KEY here \n Begins with ${acceptedAlgorithms
+            .map(el => "'" + el + "'")
+            .join(', ')}`}
+        />
+      </Form.Item>
+      ​
+      <Form.Item>
+        <Row justify="end">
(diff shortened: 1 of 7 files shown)


Ale142:
The SSH key validation checks that the format of the key follows the specification rules of OpenSSH.

As It said in the code, the name field is required and it is appended to the key as a comment.

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+            message: 'Field required',
+          },
+        ]}
+      >
+        <Input />
+      </Form.Item>
+
+      <Form.Item
+        name="key"
+        label="Key"
```
GabriFila, line 76: I would add "_Public_" also here

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+        />
+      </Form.Item>
+
+      <Form.Item>
+        <Row justify="end">
+          <Button type="default" htmlType="button" onClick={cancelForm}>
+            Cancel
+          </Button>
+          <Button className="ml-2" type="primary" htmlType="submit">
+            Save
```
GabriFila, line 104: This button should be disabled when the form is incorrect or incomplete

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+    return false;
+  };
+
+  const submitForm = (fields: { name: string; key: string }) => {
+    /* the name field is appended to the key */
+    const comment = fields.key.split(' ')[2];
+    if (comment) fields.key = fields.key + ':' + fields.name;
+    else fields.key = fields.key + ' ' + fields.name;
```
GabriFila, line 45: Use template string notation

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (key === undefined || key.length === 0) return false;
+    const result = key.split(' ');
+
+    if (result.length === 2 || result.length === 3) {
+      if (acceptedAlgorithms.includes(result[0]) && result[1] !== '') {
+        return true;
+      }
+    }
+    return false;
+  };
```
QcFe, line 40: There's something I'd like to review about validation, please see Slack

GabriFila (approved):

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+    setValidForm(false);
+    if (!key) throw new Error('Key field is mandatory');
+    const result = key.split(/\s+/); // split regardless the amount of spaces
+    if (result.length !== 2 && result.length !== 3)
+      throw new Error('Invalid key format');
+    if (result[1] === '') throw new Error('Empty SSH key');
+    if (!acceptedAlgorithms.includes(result[0]))
+      throw new Error('Unrecognized SSH key format');
+
+    setValidForm(true);
```
QcFe, line 39: I think setValid shouldn't be called here, otherwise the form would be set as valid also if no name is provided

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+    form.resetFields();
+  };
+
+  return (
+    <Form
+      form={form}
+      labelCol={{ span: 4 }}
+      wrapperCol={{ span: 24 }}
+      onFinish={submitForm}
+      onChange={handleChange}
```
QcFe, line 75: I don't remember if there was any issue with using `onValuesChange`, it seems from documentation that it's not really advisable to use onChange... Also `onValuesChange` could make code much cleaner...

QcFe (changes requested):
There's a behaviour that's still not correct

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const result = key.split(/\s+/); // split regardless the amount of spaces
+    if (result.length !== 2 && result.length !== 3)
+      throw new Error('Invalid key format');
+    if (result[1] === '') throw new Error('Empty SSH key');
+    if (!acceptedAlgorithms.includes(result[0]))
+      throw new Error('Unrecognized SSH key format');
+  };
+
+  // Used to disable Save button
+  const handleChange = async () => {
```
QcFe, line 40:
```suggestion
  const handleChange = async ({ name, key }: { name: string; key: string }) => {
```
If I'm not wrong, onValuesChange should receive parameters similarly to the submit. this should make the code cleaner and avoid the need to use getFieldValue

Ale142: I tried passing the `name` and `key` parameters to `handleChange` but I noticed some strange behavior. When `onValuesChange` (and handleChange) is triggered only the field that actually changes is set, the other is not. For example, if you write the key only the `key` parameter contains the typed value, `name` remains `undefined`.

QcFe: By looking at the code we have 
```ts
export interface Callbacks<Values = any> {
  //...
  onValuesChange?: (changedValues: any, values: Values) => void
}
```
so this means that you'd be able to do 
```ts
  const handleChange = async (_, { name, key }: { name: string; key: string }) => {
    //...
  }
```
(notice the first parameter that's just an underscore)
Have a try like this if you could

Ale142: Yes! Not it works thanks a lot :)

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error('Unrecognized SSH key format');
+  };
+
+  // Used to disable Save button
+  const handleChange = async (
+    _: any,
+    { name, key }: { name: string; key: string }
+  ) => {
+    try {
+      if (key !== undefined && name !== undefined) {
```
QcFe, line 45:
```suggestion
      if (key && name) {
```
Should be enough, better to test it though

frontend/src/components/accountPage/SSHKeysForm/SSHKeysForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+        await form.validateFields(['name', 'key']);
+        setValidForm(true);
+      } else if (key !== undefined) await form.validateFields(['key']);
```
QcFe, line 48:
```suggestion
      } else if (key) await form.validateFields(['key']);
```
same as above

QcFe (approved):
That's good for me. If you wanna review just that last little thing then we should be ready to merge!

QcFe:
Ok, for me it's alright. Please squash everything in a single commit and run a final deploy-staging just to be sure everything is alright, then I think we can merge.

PR #471, opened 2021-05-15
Turn off coveralls check if non-operators files are changed

GabriFila:
# Description

This PR disables the action step that sends the coveralls check to avoid having a red cross on master/PRs that do not touch `.go` files and files that are in the `operators` folder

# How Has This Been Tested?

This has been tested by temporarily changing the action trigger to a push on every branch and then making some commits that would trigger the action

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,14 @@ jobs:
           repository: "${{ needs.configure.outputs.repo-name }}"
           persist-credentials: false
 
+      - name: Check if /operators files changed
+        uses: dorny/paths-filter@v2
+        id: pathFilter
+        with:
+          filters: |
+            operators:
+              - 'operators/**'
+
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -83,6 +91,7 @@ jobs:
           make test-python
 
       - name: Send coverage
+        if: steps.pathFilter.outputs.operators == 'true'
         uses: shogo82148/actions-goveralls@v1
         with:
           path-to-profile: coverage.out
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,10 +1,6 @@
 name: Code testing
 on:
   push:
-    branches:
-    - master
-    tags:
-    - "v*"
```
giorio94, line 7: Should be restored

.github/workflows/test.yml
```diff
@@ -55,6 +51,16 @@ jobs:
...
 
+      - name: Check if change in /operators or in any .go file
+        uses: dorny/paths-filter@v2
+        id: pathFilter
+        with:
+          filters: |
+            go:
+              - '**.go'
+            operators:
+              - 'operators/**'
```
giorio94, line 62: The operators filter should be enough, as we do not have go files outside that folder (and they would not be tested anyhow).

giorio94 (commented):
Thanks @GabriFila for this. 

Just some minor comments, and you should also drop the test commits

QcFe:
Is the storybook-related JS file related to this PR?

.github/workflows/test.yml
```diff
@@ -55,6 +55,14 @@ jobs:
           repository: "${{ needs.configure.outputs.repo-name }}"
           persist-credentials: false
 
+      - name: Check if change in /operators or in any .go file
```
giorio94, line 58: Nit. The name should be updated

giorio94 (approved):

GabriFila:
> Is the storybook-related JS file related to this PR?

No

QcFe (approved):

PR #474, opened 2021-05-22
Component Box - Fixes

cannarelladev:
# Description

With this minor fix I solved some prolems with the header element disposition. Until now when a user tried to insert only one element on the left o right  without other content it was rendered on center. Now it's possible to also use only "left" and "right" properties of object "header" and fix respectively the content. Moreover i removed all that div blocks and their rigorous Style Classes that constrain content disposition (i.e. vertical and horizonal alignment), so now the Component is more customizable. Latest, i add the faculty to add a predefined min height for header or footer when the user doesn't intert "header" and "footer" content. This is usefull when a dynamic content (with scrollbar) or any other element hides rounded corner.

Fixes # (issue)

- Header content disposition
- Removed rigorous disposition Style Classes
- Added innerHeader and innerFooter properties

### DIFF
--- a/frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.less
+++ b/frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.less
@@ -1,0 +1,8 @@
+// WE HAVE TO FIX THIS HARD CODED HEIGHT
+// document.documentElement.style.setProperty('--box-ext-height', `${value}px`); 
+
+.cl-templates-table{
+  min-height: 171px;  
+  height: calc(100vh - var(--box-ext-height, 380px)); //465px if we want to add a Box footer
+
+}
--- a/frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
+++ b/frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
@@ -1,22 +1,13 @@
 import { FC } from 'react';
-import { Col, Table } from 'antd';
+import { Table } from 'antd';
 import { TemplatesTableRow } from '../TemplatesTableRow';
-import { DownOutlined, RightOutlined } from '@ant-design/icons';
+import { RightOutlined } from '@ant-design/icons';
 import { useState } from 'react';
-import { WorkspaceRole } from '../../../../utils';
+import { Template, WorkspaceRole } from '../../../../utils';
+import './TemplatesTable.less';
 
 export interface ITemplatesTableProps {
-  templates: Array<{
-    id: string;
-    name: string;
-    gui: boolean;
-    instances: Array<{
-      id: number;
-      name: string;
-      ip: string;
-      status: boolean;
-    }>;
-  }>;
+  templates: Array<Template>;
   role: WorkspaceRole;
   editTemplate: (id: string) => void;
   deleteTemplate: (id: string) => void;
@@ -33,18 +24,8 @@ const TemplatesTable: FC<ITemplatesTableProps> = ({ ...props }) => {
     {
       title: 'Template',
       key: 'template',
-      /* eslint-disable react/no-multi-comp */
-      render: (record: {
-        id: string;
-        name: string;
-        gui: boolean;
-        instances: Array<{
-          id: number;
-          name: string;
-          ip: string;
-          status: boolean;
-        }>;
-      }) => (
+      // eslint-disable-next-line react/no-multi-comp
+      render: (record: Template) => (
         <TemplatesTableRow
           id={record.id}
           name={record.name}
@@ -55,7 +36,6 @@ const TemplatesTable: FC<ITemplatesTableProps> = ({ ...props }) => {
           deleteTemplate={deleteTemplate}
         />
       ),
-      /* eslint-enable react/no-multi-comp */
     },
   ];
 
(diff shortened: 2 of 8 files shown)


QcFe:
I fear I didn't understand what the _inner_* props of the box serve to...

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
```diff
@@ -96,35 +63,34 @@ const TemplatesTable: FC<ITemplatesTableProps> = ({ ...props }) => {
       className="w-full flex-grow flex-wrap content-between py-0 overflow-auto scrollbar"
       style={{ height: 'calc(100vh - 380px)', minHeight: '171px' }} //465px if we want to add a Box footer
```
QcFe, line 64: I'd avoid hardcoding such kind of measures. If there's no way to calculate them from here it's better to put this as props I guess... (I'm mostly referring to the _380px_)

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
```diff
@@ -96,35 +63,34 @@ const TemplatesTable: FC<ITemplatesTableProps> = ({ ...props }) => {
...
-             */
-            expandedRowRender: template => 'Running Instances',
-          }}
-        />
-      </Col>
+                <RightOutlined onClick={e => onExpand(record, e)} />
+              )
+            ) : (
+              false
+            ),
```
QcFe, line 87: Isn't the `expr && elm` "pattern" working here?

frontend/src/utils.ts
```diff
@@ -2,3 +2,14 @@ export type someKeysOf<T> = { [key in keyof T]?: T[key] };
...
+export type Template = {
+  id: string;
+  name: string;
+  gui: boolean;
+  instances: Array<{
+    id: number;
+    name: string;
+    ip: string;
+    status: boolean;
+  }>;
```
QcFe, line 14: Was wondering if making an _Instance_ interface could make things more organized...

cannarelladev: it could be a good idea, i do the changes you suggest

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
cannarelladev: it was an overview, i'll test and correct it, thanks

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
cannarelladev: actually i moved this style properties (height and min-height) to a CSS file. About the fixed value of `calc` i fear it must be a fixed value beacuse it's necessary to activate the scrollbar of the table. the final correct pixel value will be found when we insert all the components in a page with header (navbar) and footer. based on these last elements we can decide the correct value. Tell me if it could be the right way to proceed

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
cannarelladev: UPDATE: i test you change suggestion but it not work because the new expression is rendered as `0` was the result is `false`. I think i have to leave everything unchanged

QcFe: That's what I feared

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
QcFe: It could be better to use a css variable (that could be set inside a parent or something) then, or still make it settable somewhere... Otherwise can't we set it as full-height and contain it from outside?
The fact is if we have such kind of value that is decided inside the component but depends on something outside... It's not really a component...

GabriFila (approved):

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.less
```diff
@@ -0,0 +1,7 @@
+// WE HAVE TO FIX THIS HARD CODED HEIGHT
+@ext-height: 380px; //465px if we want to add a Box footer
+
+.cl-templates-table{
+  min-height: 171px;  
+  height: calc(100vh - @ext-height); 
+}
```
QcFe, line 7:
```suggestion
// document.documentElement.style.setProperty('--box-ext-height', `${value}px`);

.cl-templates-table {
  min-height: 171px;  
  height: calc(100vh - var(--box-ext-height, 380px)); 
}
```
I would rather use CSS variables so that we could easily set them on runtime through JS or css breakpoints...

QcFe (approved):
Just a minor note

frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.less
cannarelladev: i didn't know this way, so ok i agree. Do i commit and squash this fix?

QcFe: That's good for me. We should jsut see if it works correctly then for me we can merge

cannarelladev: ok, i proceed

PR #475, opened 2021-05-25
Added ModalCreateTemplate

cheina97:
Added components:

- ModalCreateTemplate

### DIFF
--- a/frontend/src/Decorators.tsx
+++ b/frontend/src/Decorators.tsx
@@ -13,7 +13,9 @@ const ModalDecorator = (Story: Story, context: StoryContext) => {
     <Layout>
       <Layout.Content>
         <Story />
-        <Skeleton />
+        <div className="m-4">
+          <Skeleton />
+        </div>
         <div className=" flex justify-center mt-12">
           <Button
             onClick={() => setShow(true)}
--- a/frontend/src/components/workspaces/ModalCreateTemplate/ModalCreateTemplate.stories.tsx
+++ b/frontend/src/components/workspaces/ModalCreateTemplate/ModalCreateTemplate.stories.tsx
@@ -1,0 +1,52 @@
+import ModalCreateTemplate, {
+  IModalCreateTemplateProps,
+  Template,
+} from './ModalCreateTemplate';
+import { Story, Meta } from '@storybook/react';
+import { ModalDecorator } from '../../../Decorators';
+
+export default {
+  title: 'Components/workspaces/ModalCreateTemplate',
+  component: ModalCreateTemplate,
+  argTypes: { onClick: { action: 'clicked' } },
+  decorators: [ModalDecorator],
+} as Meta;
+
+const TemplateStorybook: Story<IModalCreateTemplateProps> = args => (
+  <ModalCreateTemplate {...args} />
+);
+
+export const Create = TemplateStorybook.bind({});
+Create.args = {
+  submitHandler: (t?: Template) => {
+    alert(JSON.stringify(t));
+  },
+  diskInterval: { min: 1, max: 32 },
+  ramInterval: { min: 4, max: 16 },
+  cpuInterval: { min: 1, max: 4 },
+  images: [
+    {
+      name: 'Ubuntu',
+      vmorcontainer: ['Container', 'VM'],
+    },
+    {
+      name: 'Windows',
+      vmorcontainer: ['Container'],
+    },
+  ],
+};
+
+export const Modify = TemplateStorybook.bind({});
+Modify.args = {
+  ...Create.args,
+  template: {
+    name: 'Existing Template',
+    image: 'Ubuntu',
+    vmorcontainer: 'Container',
+    diskMode: false,
+    gui: true,
+    cpu: 2,
+    ram: 16,
+    disk: 24,
+  },
+};
--- a/frontend/src/components/workspaces/ModalCreateTemplate/ModalCreateTemplate.tsx
+++ b/frontend/src/components/workspaces/ModalCreateTemplate/ModalCreateTemplate.tsx
@@ -1,0 +1,452 @@
+import { useState, useEffect, FC } from 'react';
+import {
+  Modal,
+  Slider,
+  Form,
+  Input,
+  Select,
+  Checkbox,
+  Tooltip,
+  Row,
+} from 'antd';
+import Button from 'antd-button-color';
+
+const alternativeHandle = { border: 'solid 2px #1c7afdd8' };
+
+const { Option } = Select;
+
+type Image = {
+  name: string;
+  vmorcontainer: Array<Vmorcontainer>;
+};
+
+type Vmorcontainer = 'Container' | 'VM';
+
+type Template = {
+  name: string | undefined;
+  image: string | undefined;
+  vmorcontainer: Vmorcontainer | undefined;
+  diskMode: boolean;
+  gui: boolean;
+  cpu: number;
+  ram: number;
+  disk: number;
+};
+
+type Interval = {
+  max: number;
+  min: number;
+};
+
+type Valid = {
+  name: { status: string; help: string | undefined };
+  vmorcontainer: { status: string; help: string | undefined };
+};
+export interface IModalCreateTemplateProps {
+  template?: Template;
+  images: Array<Image>;
+  cpuInterval: Interval;
+  ramInterval: Interval;
+  diskInterval: Interval;
+  showModal: boolean;
+  setShowModal: (status: boolean) => void;
+  submitHandler: (t?: Template) => void;
+}
+
+const ModalCreateTemplate: FC<IModalCreateTemplateProps> = ({ ...props }) => {
+  const {
+    showModal,
+    setShowModal,
+    cpuInterval,
+    ramInterval,
+    diskInterval,
+    images,
+    template,
+    submitHandler,
+  } = props;
+
+  const [buttonDisabled, setButtonDisabled] = useState(true);
+
+  const [formTemplate, setFormTemplate] = useState<Template>({
+    name: template && template.name,
+    image: template && template.image,
+    vmorcontainer: template && template.vmorcontainer,
+    diskMode: !!template && template.diskMode,
+    gui: !!template && template.gui,
+    cpu: template ? template.cpu : cpuInterval.min,
+    ram: template ? template.ram : ramInterval.min,
+    disk: template ? template.disk : diskInterval.min,
+  });
+
+  const [valid, setValid] = useState<Valid>({
+    name: { status: 'success', help: undefined },
+    vmorcontainer: { status: 'success', help: undefined },
+  });
+
+  useEffect(() => {
+    if (
+      formTemplate.name &&
+      formTemplate.image &&
+      formTemplate.vmorcontainer &&
+      (template
+        ? template.name !== formTemplate.name ||
+          template.image !== formTemplate.image ||
+          template.vmorcontainer !== formTemplate.vmorcontainer ||
+          template.gui !== formTemplate.gui ||
+          template.diskMode !== formTemplate.diskMode ||
+          template.cpu !== formTemplate.cpu ||
+          template.ram !== formTemplate.ram ||
+          template.disk !== formTemplate.disk
+        : true)
+    )
+      setButtonDisabled(false);
+    else setButtonDisabled(true);
+  }, [formTemplate, template]);
+
+  const nameValidator = () => {
+    if (formTemplate.name === '' || formTemplate.name === undefined) {
+      setValid(old => {
+        return {
+          ...old,
+          name: { status: 'error', help: 'Please insert template name' },
+        };
+      });
+    } else {
+      setValid(old => {
+        return {
+          ...old,
+          name: { status: 'success', help: undefined },
+        };
+      });
+    }
+  };
+
+  const vmorcontainerValidator = () => {
+    if (formTemplate.vmorcontainer === undefined) {
+      setValid(old => {
+        return {
+          ...old,
+          vmorcontainer: { status: 'error', help: 'Please select' },
+        };
+      });
+    } else {
+      setValid(old => {
+        return {
+          ...old,
+          vmorcontainer: { status: 'success', help: undefined },
+        };
+      });
+    }
+  };
+
+  const [form] = Form.useForm();
+
+  const fullLayout = {
+    wrapperCol: { offset: 0, span: 24 },
+  };
+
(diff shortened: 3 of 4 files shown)


QcFe:
By now templates have just a type field (vm or container) and an image field (an image id basically, and together with it we should also know if such image has GUI support or not, and won't likely be editable). Basically the type could be used to filter out the images list (if it comes as a single list) or to switch between a list of containers and one of vms images. I'd ask @giorio94 and @GabriFila to check if that's correct.

giorio94:
> By now templates have just a type field (vm or container) and an image field (an image id basically, and together with it we should also know if such image has GUI support or not, and won't likely be editable). Basically the type could be used to filter out the images list (if it comes as a single list) or to switch between a list of containers and one of vms images. I'd ask @giorio94 and @GabriFila to check if that's correct.
> Also in case of containers it's possible to not provide storage, but probably it's not advisable to add this in the gui...
> In the meantime, I'd remove the change in App.tsx because it's irrelevant and could possibly lead to merge conflicts in case of rebases.

This part is tricky. In the long run, the user should just select an image, and the other characteristics (VM/container, GUI/No GUI) should be inferred automatically (@qcfe, this could be a task in your TODO list, even though probably with low priority). At the moment, the user needs to specify all parameters, and mainly:
* Whether the image is a VM or container-based
* Whether it has GUI or not
* Whether it is persistent or not (only in case of VMs)

cheina97:
> > By now templates have just a type field (vm or container) and an image field (an image id basically, and together with it we should also know if such image has GUI support or not, and won't likely be editable). Basically the type could be used to filter out the images list (if it comes as a single list) or to switch between a list of containers and one of vms images. I'd ask @giorio94 and @GabriFila to check if that's correct.
> > Also in case of containers it's possible to not provide storage, but probably it's not advisable to add this in the gui...
> > In the meantime, I'd remove the change in App.tsx because it's irrelevant and could possibly lead to merge conflicts in case of rebases.
> 
> This part is tricky. In the long run, the user should just select an image, and the other characteristics (VM/container, GUI/No GUI) should be inferred automatically (@QcFe, this could be a task in your TODO list, even though probably with low priority). At the moment, the creator needs to specify all parameters, and mainly:
> 
> * Whether the image is a VM or container-based
> * Whether it has GUI or not
> * Whether it is persistent or not (only in case of VMs)

I'll change the image version with container/VM and add a field persistent/not persistent on the side of environment field.
Could be a good idea to replace the SELECT component with a CHECKBOX for environment and gui fields, in the future they will have only 2 possible values right?
Adding the automatic selection of the characteristics of the selected image won't be difficult, the component implement a logic and is quite flexible.

QcFe:
@giorio94 Shouldn't containers have the option to be set to have no disk space?

giorio94:
> @giorio94 Shouldn't containers have the option to be set to have no disk space?

Yep

QcFe:
So, @cheina97, if not too tricky it would be the case to extend the minimum value of the disk value in case of containers) so that can be 0 and, in that case, maybe show an information box saying that by setting disk to 0 data loss might happen in case of crashes

cheina97:
> So, @cheina97, if not too tricky it would be the case to extend the minimum value of the disk value in case of containers) so that can be 0 and, in that case, maybe show an information box saying that by setting disk to 0 data loss might happen in case of crashes

No problem, if u like i can add a third checkbox besides the other 2 called "No disk space", with a tooltip explaining that it works only for containers and giving informations about data loss in case of crash. If this checkbox is enabled i'll disabled or hide (could be a better idea) the disk space slider.

QcFe:
Makes sense. Probably, the checkbox could replace the persistent one, so in case of vms is "persistent" and in case of containers is "enable disk" for example (then indeed, hiding the disk slider when it's deselected)

cheina97:
> Makes sense. Probably, the checkbox could replace the persistent one, so in case of vms is "persistent" and in case of containers is "enable disk" for example (then indeed, hiding the disk slider when it's deselected)

I think could be a better idea to disable the checkboxes and not to hide them. Imagine you are a user that don't know that containers and VMs have different options, if you try to select container you will see a disabled option and you will know that VMs have this option available. Maybe someone could not be aware of different options if don't see a disabled option. Tell me what you think about it or we could discuss about it in the next meeting.

giorio94:
> > Makes sense. Probably, the checkbox could replace the persistent one, so in case of vms is "persistent" and in case of containers is "enable disk" for example (then indeed, hiding the disk slider when it's deselected)
> 
> I think could be a better idea to disable the checkboxes and not to hide them. Imagine you are a user that don't know that containers and VMs have different options, if you try to select container you will see a disabled option and you will know that VMs have this option available. Maybe someone could not be aware of different options if don't see a disabled option. Tell me what you think about it or we could discuss about it in the next meeting.

Just my two cents. Why do you need two different checkboxes? Couldn't it be just one "persistent" or whatever we decide to call it, independently from the underlying environment? In the end, the user should not be even aware of the fact a given environment is a VM or a container (currently, this is not yet possible due to a lack in the specs of the imagelist resource).

cheina97:
> > > Makes sense. Probably, the checkbox could replace the persistent one, so in case of vms is "persistent" and in case of containers is "enable disk" for example (then indeed, hiding the disk slider when it's deselected)
> > 
> > 
> > I think could be a better idea to disable the checkboxes and not to hide them. Imagine you are a user that don't know that containers and VMs have different options, if you try to select container you will see a disabled option and you will know that VMs have this option available. Maybe someone could not be aware of different options if don't see a disabled option. Tell me what you think about it or we could discuss about it in the next meeting.
> 
> Just my two cents. Why do you need two different checkboxes? Couldn't it be just one "persistent" or whatever we decide to call it, independently from the underlying environment? In the end, the user should not be even aware of the fact a given environment is a VM or a container (currently, this is not yet possible due to a lack in the specs of the imagelist resource).

Maybe i misunderstood, the creator doesn't have to specify if the environment will be a VM or a container? 
The two checkboxes are used, one to specify that a VM is persistent, the other to specify whether the container will have disk space or not. 
What I meant with the last message is that not hiding the persistence checkbox when the container option is already selected may help the user to notice that VM and container have different options.
Obviously the checkboxes will be set to false and disabled if the option is not available based in the currently selected environment (i mean VM/container). If i misunderstood, sorry i'll fix it before the next meeting.

giorio94:
> > > > Makes sense. Probably, the checkbox could replace the persistent one, so in case of vms is "persistent" and in case of containers is "enable disk" for example (then indeed, hiding the disk slider when it's deselected)
> > > 
> > > 
> > > I think could be a better idea to disable the checkboxes and not to hide them. Imagine you are a user that don't know that containers and VMs have different options, if you try to select container you will see a disabled option and you will know that VMs have this option available. Maybe someone could not be aware of different options if don't see a disabled option. Tell me what you think about it or we could discuss about it in the next meeting.
> > 
> > 
> > Just my two cents. Why do you need two different checkboxes? Couldn't it be just one "persistent" or whatever we decide to call it, independently from the underlying environment? In the end, the user should not be even aware of the fact a given environment is a VM or a container (currently, this is not yet possible due to a lack in the specs of the imagelist resource).
> 
> Maybe i misunderstood, the creator doesn't have to specify if the environment will be a VM or a container?

At the moment it has (as you are correctly saying) due to technical limitations, but eventually it shouldn't (should be inferred from the selected image).

> The two checkboxes are used, one to specify that a VM is persistent, the other to specify whether the container will have disk space or not.
> What I meant with the last message is that not hiding the persistence checkbox when the container option is already selected may help the user to notice that VM and container have different options.
> Obviously the checkboxes will be set to false and disabled if the option is not available based in the currently selected environment (i mean VM/container). If i misunderstood, sorry i'll fix it before the next meeting.

Not clear to me why you need two checkboxes, one "persistent" for a VM, and one "has disk space" for the containers. In the end, they have the same meaning (i.e., create a persistent disk), and both enable the slider to select the amount of disk space (although with different ranges based on the type of environment.

QcFe:
It's basically the same thing I suggested, to have just one single checkbox (whose label is "persistent" for vms and "has disk" for containers). I think showing useless controls (and justifying why they're disabled) can be confusing also because as Marco said, the type of environment will depend on the image itself. This because, for example, some applications can run inside containers with no problems (and there shouldn't be actual reasons to have them into vms) while others can't be run inside containers, thus has to be run inside vms. So it doesn't make much sense to give such kind of choice for the same application

cheina97:
> Not clear to me why you need two checkboxes, one "persistent" for a VM, and one "has disk space" for the containers. In the end, they have the same meaning (i.e., create a persistent disk), and both enable the slider to select the amount of disk space (although with different ranges based on the type of environment.

I understood from [this message](https://github.com/netgroup-polito/CrownLabs/pull/475#issuecomment-853748084) that persistent or not persistent have to be a choice only for VMs and not for containers. 
So both containers and VMs can be persistent ?

cheina97:
> It's basically the same thing I suggested, to have just one single checkbox (whose label is "persistent" for vms and "has disk" for containers). I think showing useless controls (and justifying why they're disabled) can be confusing also because as Marco said, the type of environment will depend on the image itself. This because, for example, some applications can run inside containers with no problems (and there shouldn't be actual reasons to have them into vms) while others can't be run inside containers, thus has to be run inside vms. So it doesn't make much sense to give such kind of choice for the same application

I agree, what I said about checkboxes no longer makes sense if there can be no VM / container choice

giorio94:
> > Not clear to me why you need two checkboxes, one "persistent" for a VM, and one "has disk space" for the containers. In the end, they have the same meaning (i.e., create a persistent disk), and both enable the slider to select the amount of disk space (although with different ranges based on the type of environment.
> 
> I understood from [this message](https://github.com/netgroup-polito/CrownLabs/pull/475#issuecomment-853748084) that persistent or not persistent have to be a choice only for VMs and not for containers.
> So both containers and VMs can be persistent ?

You are right, that comment was misleading. In the stricter sense, only VMs can be persistent, but the same concept can extended to containers with/without disk (which makes them persistent/non persistent)

giorio94:
> It's basically the same thing I suggested, to have just one single checkbox (whose label is "persistent" for vms and "has disk" for containers). I think showing useless controls (and justifying why they're disabled) can be confusing also because as Marco said, the type of environment will depend on the image itself. This because, for example, some applications can run inside containers with no problems (and there shouldn't be actual reasons to have them into vms) while others can't be run inside containers, thus has to be run inside vms. So it doesn't make much sense to give such kind of choice for the same application

To me, the label could be also persistent in both cases

giorio94:
> > It's basically the same thing I suggested, to have just one single checkbox (whose label is "persistent" for vms and "has disk" for containers). I think showing useless controls (and justifying why they're disabled) can be confusing also because as Marco said, the type of environment will depend on the image itself. This because, for example, some applications can run inside containers with no problems (and there shouldn't be actual reasons to have them into vms) while others can't be run inside containers, thus has to be run inside vms. So it doesn't make much sense to give such kind of choice for the same application
> 
> I agree, what I said about checkboxes no longer makes sense if there can be no VM / container choice

Just for clarity, this is in the long term. Currently, the checkbox VM/container is still needed.

cheina97:
> > > Not clear to me why you need two checkboxes, one "persistent" for a VM, and one "has disk space" for the containers. In the end, they have the same meaning (i.e., create a persistent disk), and both enable the slider to select the amount of disk space (although with different ranges based on the type of environment.
> > 
> > 
> > I understood from [this message](https://github.com/netgroup-polito/CrownLabs/pull/475#issuecomment-853748084) that persistent or not persistent have to be a choice only for VMs and not for containers.
> > So both containers and VMs can be persistent ?
> 
> You are right, that comment was misleading. In the stricter sense, only VMs can be persistent, but the same concept can extended to containers with/without disk (which makes them persistent/non persistent)

Ok thanks, in the meanwhile i talked also with @QcFe and now it's all clear. Thanks for yours time and sorry for the inconvenience

QcFe:
> To me, the label could be also persistent in both cases
It wouldn't be too similar however functionally and technically. To enable attaching the PVC by now we're checking if the disk size is more than 0, regardless the persistent flag. And by now containers can't be stopped (like persistent vms). Probably it could be the case to unify such kind of behaviour at this point...

giorio94:
> > To me, the label could be also persistent in both cases
> 
> It wouldn't be too similar however functionally and technically. To enable attaching the PVC by now we're checking if the disk size is more than 0, regardless the persistent flag. And by now containers can't be stopped (like persistent vms). Probably it could be the case to unify such kind of behaviour at this point...

Yes, wouldn't focus too much on the current implementation to design the GUI. And I agree the behavior should be uniformed as much as possible during the instance-operator refactoring.

GabriFila (approved):

giorio94:
Was looking at the storybook, overall looks ok to me. I just noticed a couple of minor problems:
* The checkboxes seems not centered.
* The behavior of the persistent one is inverted (the disk slider should be enabled when checked).

cheina97:
> Was looking at the storybook, overall looks ok to me. I just noticed a couple of minor problems:
> 
> * The checkboxes seems not centered.
> * The behavior of the persistent one is inverted (the disk slider should be enabled when checked).

I made the changes. Tell me if u like the mobile view. Could be a good idea to maintain the old positioning (left justified) in mobile view?

giorio94:
> > Was looking at the storybook, overall looks ok to me. I just noticed a couple of minor problems:
> > 
> > * The checkboxes seems not centered.
> > * The behavior of the persistent one is inverted (the disk slider should be enabled when checked).
> 
> I made the changes. Tell me if u like the mobile view. Could be a good idea to maintain the old positioning (left justified) in mobile view?

Personally, I think that the centering positioning is fine also in the mobile view. I would just make the checkbox be aside the corresponding label, as it seems there is enough space. Additionally, I would keep the persistent one enabled (i.e. not greyed out) from the beginning, just for consistency with the other.

cheina97:
> > > Was looking at the storybook, overall looks ok to me. I just noticed a couple of minor problems:
> > > 
> > > * The checkboxes seems not centered.
> > > * The behavior of the persistent one is inverted (the disk slider should be enabled when checked).
> > 
> > 
> > I made the changes. Tell me if u like the mobile view. Could be a good idea to maintain the old positioning (left justified) in mobile view?
> 
> Personally, I think that the centering positioning is fine also in the mobile view. I would just make the checkbox be aside the corresponding label, as it seems there is enough space. Additionally, I would keep the persistent one enabled (i.e. not greyed out) from the beginning, just for consistency with the other.

Ok, i changed it.

giorio94 (approved):

PR #477, opened 2021-05-27
Try to make blender aware of available cpu cores

QcFe:
# Description

Blender automatically gets the number of available cpu cores to decide how many threads spawn for rendering. This PR includes:
- Environment variables for the containers in order to let them know their actual cpu limits and requests
- Blender startup script to set those values on scene load

# How Has This Been Tested?

- [x] Local blender setup
- [x] Blender on CrownLabs

### DIFF
--- a/operators/pkg/instance-controller/container_logic.go
+++ b/operators/pkg/instance-controller/container_logic.go
@@ -223,6 +223,23 @@ func buildContainerInstanceDeploymentSpec(
 				environment.Resources.ReservedCPUPercentage,
 				environment.Resources.Memory,
 			),
+			Env: []v1.EnvVar{{
+				ValueFrom: &v1.EnvVarSource{
+					ResourceFieldRef: &v1.ResourceFieldSelector{
+						ContainerName: name,
+						Resource:      "requests.cpu",
+					},
+				},
+				Name: "CROWNLABS_CPU_REQUESTS",
+			}, {
+				ValueFrom: &v1.EnvVarSource{
+					ResourceFieldRef: &v1.ResourceFieldSelector{
+						ContainerName: name,
+						Resource:      "limits.cpu",
+					},
+				},
+				Name: "CROWNLABS_CPU_LIMITS",
+			}},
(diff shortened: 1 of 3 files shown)


QcFe:
@giorio94 I tested it on the staging env and it seems to work, if you want to check it out I made a testing template on the comp-anim course...
If it seems good to you I think it could kind of be mergeable...

operators/pkg/instance-controller/container_logic.go
```diff
@@ -223,6 +223,23 @@ func buildContainerInstanceDeploymentSpec(
...
+				},
+				Name: "CROWNLABS_CPU_REQUESTS",
+			}, {
+				ValueFrom: &v1.EnvVarSource{
+					ResourceFieldRef: &v1.ResourceFieldSelector{
+						ContainerName: name,
+						Resource:      "limits.cpu",
+					},
+				},
+				Name: "CROWNLABS_CPU_LIMTIS",
```
giorio94, line 241:
```suggestion
				Name: "CROWNLABS_CPU_LIMITS",
```

provisioning/containers/blender/autolimits.py
```diff
@@ -0,0 +1,21 @@
+import os
+import bpy
+from bpy.app.handlers import persistent
+
+
+@persistent  # Keep across file reloads
+def cpulimits_setter(_):  # An argument has to be present for handlers
+    try:
+        maxThreads = int(float(os.environ['CROWNLABS_CPU_LIMTIS']))  # Round possible floats
```
giorio94, line 9: Would round up, to avoid issues in case of limits <1CPU (and avoid wasting available resources if limits are not integers).

giorio94, line 9:
```suggestion
        maxThreads = int(float(os.environ['CROWNLABS_CPU_LIMITS']))  # Round possible floats
```

giorio94 (approved):
/lgtm

Just a couple of minor comments

GabriFila (approved):

PR #478, opened 2021-05-30
Active instances table and relative components

Nepec:
# Description

This PR includes the intances table with all associated components. This is the last table of my nesting chain and each row consists in two columns, one for the icons and one for the content itself.  
The icons are:
- Intance type: GUI or CLI
- Status
- A instance specific avatar

The contents are further divided in three elements:
- A heading
- An info icon
- The available actions.

When in mobile view the icons column collapses leaving only the content for better readabilty. Also the secodary action buttons collapse behind a *more* icon.

# ToDo
- [ ] Differentiate managed/personal instances layout. Currently the only differences are in the heading but they should also be in:
  - Icons: the managed instances don't need the type icon and the avatar,
  - Info icon: managed should not have one,
  - Actions: specific per instance type
- [ ] Info button dropdown
- [ ] Add Files button dropdown
- [ ] Destory button alert
- [ ] Button Hooks
- [ ] Add responsive ellipses to text fields
- [ ] Review file structure (maybe?)
- [ ] Better stories documentation

# Notes

Added a new prop to `NestedTables` to differentiate the nested status from the manager one. The former has layout purposes in the nesting while the latter has actual semantic meaning and needs to be used by the instances too.


frontend/src/components/activePage/NestedTables/NestedTables.tsx
```diff
@@ -23,25 +26,26 @@ export interface INestedTablesProps {
   workspaces: Array<IWorkspace>;
   templates: Array<ITemplate>;
   isManager: boolean;
+  nested: boolean;
   destroyAll: () => void;
 }
 
 type rowType = IWorkspace | ITemplate;
 
 const NestedTables: FC<INestedTablesProps> = ({ ...props }) => {
```
GabriFila, line 35: Don't know if it makes sense: since it takes a `nested` prop, maybe call it `NestableTable`? What do you think?

frontend/src/components/activePage/Instances/InstanceIcons/InstanceIcons.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export interface IInstanceIconsProps {
+  isGUI: boolean;
+  phase: 'ready' | 'creating' | 'failed' | 'stopping' | 'off';
+}
+
+const InstanceIcons: FC<IInstanceIconsProps> = ({ ...props }) => {
+  const { isGUI, phase } = props;
+
+  const chooseStatusIcon = () => {
```
GabriFila, line 20: Just a suggestion, you could define a object (outside the component) with:
- key: phase
- value: component corresponding to the phase
And then instead of calling the function you take the value from the object

frontend/src/components/activePage/Instances/InstanceIcons/InstanceIcons.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <>
+          <Title>Workpaces/Templates table</Title>
+          <Description />
+        </>;
+      },
+    },
+  },
+};
+
+const defaultArgs: someKeysOf<IInstanceIconsProps> = {
```
GabriFila, line 21: Add the different possible icons/phases

frontend/src/components/activePage/Instances/InstanceContent/ManagedInstanceHeading.tsx
```diff
@@ -0,0 +1,25 @@
...
+  tenantId: string;
+  tenantDisplayName: string;
+  displayName: string;
+}
+
+const ManagedInstanceHeading: FC<IManagedInstanceHeadingProps> = ({
+  ...props
+}) => {
+  const { tenantId, tenantDisplayName, displayName } = props;
+  const { Text } = Typography;
```
GabriFila, line 14: I believe you can import directly the `Text` component

frontend/src/components/activePage/Instances/InstanceContent/ManagedInstanceHeading.tsx
```diff
@@ -0,0 +1,25 @@
+import { FC } from 'react';
+import { Typography } from 'antd';
+
+export interface IManagedInstanceHeadingProps {
+  tenantId: string;
+  tenantDisplayName: string;
+  displayName: string;
+}
+
+const ManagedInstanceHeading: FC<IManagedInstanceHeadingProps> = ({
```
GabriFila, line 10: If you need a different component, create a new folder (I guess you referred tyo this in the PR description)

frontend/src/components/activePage/NestedTables/NestedTables.tsx
Nepec: Sure, It would be more specific. 
Actually I don't really like the prop name "nested" because also the `InstancesTable` is nested, it just is a different component with a different shape. Basically what the `nested` prop does is it lets me decide whether to display the same table a second time or whether to display a different, yet still nested, table. Maybe I should just be explicit and call it `recurring` or something, any suggestions?

frontend/src/components/activePage/NestedTables/NestedTables.tsx
```diff
@@ -4,6 +4,9 @@ import { FC, useState } from 'react';
 import { Table } from 'antd';
 import { CaretRightOutlined } from '@ant-design/icons';
 import RowHeading from '../RowHeading/RowHeading';
+import InstancesTable from '../Instances/InstancesTable/InstancesTable';
+import { IInstance } from '../Instances/InstancesTable/InstancesTable';
+import { instances } from '../tempData';
```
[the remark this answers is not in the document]
Nepec: I wanted to isolate the `InstancesTable` component as much as possible by passing it only the instances it had to display, therefore I need the whole instances array in order to filter it for the template name.

I could also forego this approach completely and pass the template name to the `InstancesTable`, import the instances array in that component directly and filter it in the component itself, what do you think is best?

frontend/src/components/activePage/Instances/InstanceContent/ManagedInstanceHeading.tsx
Nepec: `import { Text } from 'antd';`

Does not work so I am afraid not :/

frontend/src/components/activePage/Instances/InstanceIcons/InstanceIcons.stories.tsx
Nepec: Do you mean as different Stories?

PR #485, opened 2021-06-23
UserList and UserListForm

Ale142:
# Description

This PR includes the UserList component for both add many and single user with the UserListForm modal.

## TODO:
- [x] CSV parsing & validation
- [x] Fix UserListForm validation


luco5826:
We have updated this PR including the [PapaParse](https://www.papaparse.com/) CSV parsing library and fixing the "Add one user" form validation

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <Form.Item
+        name="userid"
+        label="User ID"
+        validateTrigger="onBlur"
+        rules={[
+          {
+            required: true,
+            message: 'Field required',
+          },
+        ]}
```
QcFe, line 52:
```suggestion
        rules={[{
          required: true,
          message: 'Field required',
        }]}
```
I'd keep it compact wherever possible, JSX is already super intendous on its own (provided eslint doesn't complain)...
(same thing for the other fields)

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+            },
+          },
+        ]}
+      >
+        <Input type="email" />
+      </Form.Item>
+      ​
+      <Form.Item>
+        <Row justify="end">
+          <Button type="default" htmlType="button" onClick={cancelForm}>
```
QcFe, line 101: What's `htmlType="button"` for? 🤔

QcFe (commented):
Not bad by now. Several things to be reviewed though.
Also I'd add a search bar/button on top (right) and pagination.
In case of pagination, the buttons for adding users could be probably put in the same row of the pagination (in those case maybe leave the buttons on the left, to have pagination on the right)

QcFe:
Here is an example of csv file: be careful because it could be weird and not get parsed:
```csv
MATRICOLA,COGNOME - (*) Inserito dal docente,NOME,EMAIL,COD.INS. E CREDITI,CDS STUDENTE,FREQUENZA,ESAME,ATHOME,SDSS,INFO MOBILITA,STUDENTE ALL'ESTERO,INS IN L.A.,ANNO ISCRIZ.,CONDIZIONE
003456,MALELI(*),GOFFREDO,s123456@studenti.uniba.it,98ABCDE (6),CASDA3,Frequentato nel 1847/1848,Sostenuto il 01/02/1848,28, - , - , - ,N,N,1840/1841,RINUNCIA
212121,ACASO,STUDENTE,s212121@studenti.polito.it,98ABCDE (6),INFZT3,Frequentato nel 2020/2021, - , - , - , - , - ,N,N,2019/20,ATTIVO
```

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
```diff
@@ -0,0 +1,118 @@
+import { Input, Form, Button, Row } from 'antd';
+import { FC, useState } from 'react';
+import { User } from '../UserList/UserList';
+
+export interface IUserListFormProps {
+  onAddUser: (newUser: User) => void;
+  onCancel: () => void;
+}
+const regexEmail = /^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+(.[a-zA-Z0-9-])+$/;
```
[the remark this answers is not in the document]
Ale142: Yes, I found the way in which antd does the email validation

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <Button onClick={() => setshowUserListModal(true)} className="m-1">
+          Add one
+        </Button>
+        <Upload
+          name="file"
+          beforeUpload={() => {
+            // By default Ant tries to POST the file to an endpoint
+            // This workaround prevents it
+            return false;
+          }}
```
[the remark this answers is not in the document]
Ale142: Why do you want to remove the list of uploaded files? I think it is useful for the user to see the name of the uploaded file ..

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
Ale142: I tried to keep the syntax more compact but eslint complained..

Ale142:
> Not bad by now. Several things to be reviewed though.
> Also I'd add a search bar/button on top (right) and pagination.

Instead of adding a single search bar I opted for [this](https://ant.design/components/table/#components-table-demo-custom-filter-panel) solution that I found in the doc, let me know what do you think.
Also I added the possibility to sort the columns by **userid**, **name**, **surname**. I thought It was a good thing, again let me know what you think. ( I followed [this](https://ant.design/components/table/#components-table-demo-multiple-sorter) example).

> In case of pagination, the buttons for adding users could be probably put in the same row of the pagination (in those case maybe leave the buttons on the left, to have pagination on the right)

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
QcFe: Since this is used more than once though you could make something like `const requiredFieldRuleSet = [{...}];` before the jsx in order to reuse it

QcFe:
> Instead of adding a single search bar I opted for [this](https://ant.design/components/table/#components-table-demo-custom-filter-panel) solution that I found in the doc, let me know what do you think.
> Also I added the possibility to sort the columns by **userid**, **name**, **surname**. I thought It was a good thing, again let me know what you think. ( I followed [this](https://ant.design/components/table/#components-table-demo-multiple-sorter) example).

I think this is alright, even if I personally don't like it too much since it's not that immediate, but since it's antd's default I guess it can stay that way. 
What about the pagination instead?
(tip: when you're done with a comment in the code you can click "resolve" so we know it's done, leave it open if you think there's more to say/do though)

Ale142:
> > Instead of adding a single search bar I opted for [this](https://ant.design/components/table/#components-table-demo-custom-filter-panel) solution that I found in the doc, let me know what do you think.
> > Also I added the possibility to sort the columns by **userid**, **name**, **surname**. I thought It was a good thing, again let me know what you think. ( I followed [this](https://ant.design/components/table/#components-table-demo-multiple-sorter) example).
> 
> I think this is alright, even if I personally don't like it too much since it's not that immediate to see those search buttons, but since it's antd's default (and it's already made) I guess it can stay that way. 
> What about the pagination instead?
I think pagination is automatic when you upload a number of users
> (tip: when you're done with a comment in the code you can click "resolve" so we know it's done, leave it open if you think there's more to say/do though)
Sorry, forgot about that, now I think is fixed

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,340 @@
...
+          key="userid"
+          filterIcon={<SearchOutlined />}
+          onFilter={(value: any, record: any) =>
+            record['userid']
+              ? record['userid']
+                  .toString()
+                  .toLowerCase()
+                  .includes(value.toLowerCase())
+              : ''
+          }
```
QcFe, line 104: Please avoid replicating the code. There are several duplicated lambdas that can be generalized.
In this case you could do something like:
```ts
function makeFilter(field: string) {
  return (val: any, record: any) => 
    record[field] ? record[field].toString().toLowerCase().includes(val.toLowerCase()) : '';
}
```
then do `onFilter={makeFilter('userid')}`
Note: makeFilter is a function that returns a function.

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,340 @@
...
+            selectedKeys,
+            confirm,
+            clearFilters,
+          }) => (
+            <div style={{ padding: 8 }}>
+              <Input
+                placeholder="Search by Userid"
+                onChange={e => {
+                  setSelectedKeys(e.target.value ? [e.target.value] : []);
+                }}
```
QcFe, line 116: This one is duplicated if I'm not wrong as well. 
You could extract this lambda as a regular function, then call it here.

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,340 @@
...
+                    : '';
+                  setSearchText(currValue);
+                }}
+              />
+              <Space>
+                <Button
+                  onClick={() => {
+                    if (clearFilters) clearFilters();
+                    setSearchText('');
+                  }}
```
QcFe, line 131: This one can be extracted as well to avoid duplication.

QcFe: You could even make a sub component or just make a variable containing the button, in order to reuse it every time is needed.

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+      <Form.Item
+        name="userid"
+        label="User ID"
+        validateTrigger="onBlur"
+        rules={[
+          {
+            required: true,
+            message: 'This field cannot be empty',
+          },
+        ]}
```
QcFe, line 59: Again here, to avoid duplication, you could make a `const requiredFieldRule = { required: true, ... }` so that you can just do `rules={[requiredFieldRule]}` for example.

QcFe (changes requested):
There's quite a bit of duplicated code. I left some indications but I'd ask you to check since it seems there's quite a good margin of optimization, duplicated core is harder to be maintained. 
Take also into account if it's not the case to generalize certain components. I don't think it's always required to make a story for the little sub components (I was thinking especially to the column one here for example). Don't know if @GabriFila has any other advice on this.

frontend/src/components/accountPage/UserList/ResetButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <Button
+      onClick={() => {
+        if (clearFilters) clearFilters();
+        setSearchText('');
+      }}
+      size="small"
+      style={{ width: 90 }}
+    >
+      {children}
```
QcFe, line 19: You could even directly write "Reset" here I maybe...

frontend/src/components/accountPage/UserListForm/UserListForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+            type: 'email',
+            message: 'This is not a valid E-mail!',
+          },
+
+          requiredFieldRule,
+        ]}
+      >
+        <Input type="email" />
+      </Form.Item>
+      ​
```
QcFe, line 93: I'd remove empty newlines if possible

QcFe (approved):
Looks good to me, I'd ask you to squash into a single commit. I'll have another brief check after the squash then I think we could merge.

frontend/src/components/accountPage/UserList/UserList.tsx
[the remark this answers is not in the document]
QcFe: Mostly it's because it's ugly as it is now, but since you'd basically upload a file at a time, you still see the results of the insertion.

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    Papa.parse<User>(fileInfo.file, {
+      header: true,
+      skipEmptyLines: true,
+      complete: (result, _) => {
+        for (const line of result.data) {
+          if (!line.email || !line.name || !line.surname || !line.userid) {
+            setFileError(
+              'Invalid file format, must be <userid, name, surname, email>'
+            );
```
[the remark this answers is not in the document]
QcFe: I guess we could leave this here for now, however we'll have to properly handle this. Resolve the conversation if it's good for you.

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+export interface IUserListProps {
+  onAddUser: (users: User[]) => void;
+}
+
+const UserList: FC<IUserListProps> = props => {
+  const [showUserListModal, setshowUserListModal] = useState(false);
+  const [users, setUsers] = useState<User[]>([]);
+  const [fileError, setFileError] = useState<string>('');
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
QcFe, line 40: Is this still needed?

frontend/src/components/accountPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,304 @@
...
+                  selectedKeys,
+                  setSearchText
+                )}
+              />
+              <Space>
+                <ResetButton
+                  clearFilters={clearFilters}
+                  setSearchText={setSearchText}
+                >
+                  Reset
```
QcFe, line 251: Remove this text (check other occurrences) since it is already included in the button itself

QcFe:
Lastly please check back all the conversations and resolve them if there's nothing else to do

PR #487, opened 2021-06-28
Openapi filter

Ralls0:
# Description

This change filters the OpenApi so that the graphQL schema generates from it results much more slender.


GabriFila (approved):
LGTM, only one thing, please rename `crownlabsOas` to `targetOas`

giorio94 (approved):

PR #488, opened 2021-06-30
Fix for open api filter 

Ralls0:
# Description

Fix oas filter

### DIFF
--- a/qlkube/src/decorateOpenapi.js
+++ b/qlkube/src/decorateOpenapi.js
@@ -2,6 +2,16 @@ const { apiGroups } = require('./apiGroups');
 
 const k8s = 'io.k8s.apimachinery';
 
+/**
+ *
+ * @param {*} oas
+ * This function filters the OpenApi made from k8s API.
+ * Starting from a list of paths,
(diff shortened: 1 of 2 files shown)


giorio94 (approved):
/lgtm

qlkube/src/decorateOpenapi.js
```diff
@@ -2,14 +2,24 @@ const { apiGroups } = require('./apiGroups');
...
+
 function decorateOpenapi(oas) {
   let result = true;
   let keys = Object.keys(oas.paths);
   let removeKeys = keys.filter(key => {
     result = true;
     const paths = apiGroups.paths;
     paths.forEach(path => {
-      result &&= key.indexOf(path) == -1;
+      result = result && key.indexOf(path) === -1;
```
QcFe, line 22:
```suggestion
      result = result && !key.includes(path);
```
I'd suggest something similar to increase readability, also below

QcFe (approved):
Just a minor comment

qlkube/src/decorateOpenapi.js
QcFe: (to be tested clearly)

PR #489, opened 2021-07-01
Client configuration for graphQL

Ralls0:
# Description

Codegen configuration and Apollo-Client configuration.


frontend/public/config.js
```diff
@@ -9,3 +9,4 @@ window.REACT_APP_CROWNLABS_OIDC_PROVIDER_URL = undefined;
 window.REACT_APP_CROWNLABS_OIDC_CLIENT_ID = undefined;
 window.REACT_APP_CROWNLABS_OIDC_CLIENT_SECRET = undefined;
 window.REACT_APP_CROWNLABS_OIDC_REDIRECT_URI = undefined;
+window.REACT_APP_CROWNLABS_GRAPHQL_URL = undefined;
```
giorio94, line 12: Since you are adding a constant here, it should also be included in the helm configuration in the following locations (with an example value):
* https://github.com/netgroup-polito/CrownLabs/blob/8e73edf8fbd7fe67e528a508f221fc811dbb52b2/frontend/deploy/frontend-app/values.yaml#L17
* https://github.com/netgroup-polito/CrownLabs/blob/8e73edf8fbd7fe67e528a508f221fc811dbb52b2/deploy/crownlabs/values.yaml#L31
* https://github.com/netgroup-polito/CrownLabs/blob/8e73edf8fbd7fe67e528a508f221fc811dbb52b2/frontend/deploy/frontend-app/templates/configmap.yaml#L9

@GabriFila IIUC, we could completely drop the apiServer entry and replace that with the graphql endpoint, since there are no more direct interactions. Am I correct?

frontend/src/graphql-components/tenant/tenant.query.graphql
```diff
@@ -0,0 +1,23 @@
...
+        workspaceRef {
+          name
+          namespace
+        }
+      }
+    }
+    metadata {
+      name
+      annotations
+      labels
```
giorio94, line 20: I think these could be dropped. @GabriFila WDYT?

frontend/src/generated-types.tsx
```diff
@@ -0,0 +1,2199 @@
+import { gql } from '@apollo/client';
```
giorio94, line 1: Would mark this file as generated: https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/customizing-how-changed-files-appear-on-github

frontend/src/index.tsx
```diff
@@ -1,11 +1,42 @@
...
+      ...headers,
+      authorization: token ? `Bearer ${token}` : '',
+    },
+  };
+});
+
+const client = new ApolloClient({
+  link: authLink.concat(httpLink),
+  cache: new InMemoryCache(),
+});
```
giorio94, line 33: I would personally move this in its own file to maintain the index clean, but I am unsure whether it is a good practice.

giorio94 (commented):
A couple of minor comments inline.

frontend/src/index.tsx
Ralls0: @GabriFila  WDYT? 
To be honest, in the official [doc](https://www.apollographql.com/docs/react/networking/advanced-http-networking/) there is no indication of where is the best place to insert them.
But following this [guide](https://www.howtographql.com/react-apollo/1-getting-started/) I have seen that all materials are into the `index.js`.

QcFe:
At a first sight it looks good.
I'm not sure we need the _tsx_ file extension on those couple of files, it should be used only when those files are going to contain JSX code, @GabriFila correct me if I'm wrong. 
Also don't know if it's the case to add the missing newlines at the end of file just to get rid of the github's _no-entry_ symbol...

giorio94 (approved):

frontend/src/graphql-components/apolloClientSetup.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const authLink = setContext((_, { headers }) => {
+  const token = TOKEN;
+  return {
+    headers: {
+      ...headers,
+      authorization: token ? `Bearer ${token}` : '',
+    },
+  };
+});
```
QcFe, line 19: I'm not really understanding what's going on here, what's that empty constant token for?

GabriFila: I guess it is temporary, for development purposes, because we don't have auth in the app, hence no way to obtain the token

GabriFila (approved):

PR #490, opened 2021-07-02
instance-controller: remove oauth2-proxy creation

giorio94:
# Description

This PR removes the creation of the oauth2-proxy objects from the instance operator. It assumes the existence of a centralized oauth2-proxy deployment and configures the appropriate ingress annotations based on the URL specified as input parameter.
Finally, it removes a set of CORS-related annotation which appear to be no longer required.

# How Has This Been Tested?

- [x] Unit tests
- [x] Staging environment

### DIFF
--- a/deploy/crownlabs/Chart.lock
+++ b/deploy/crownlabs/Chart.lock
@@ -8,9 +8,12 @@ dependencies:
 - name: frontend-storybook
   repository: file://../../frontend/deploy/frontend-storybook
   version: 0.1.0
+- name: qlkube
+  repository: file://../../qlkube/deploy/qlkube
+  version: 0.1.0
 - name: instance-operator
   repository: file://../../operators/deploy/instance-operator
-  version: 0.1.0
+  version: 0.1.1
 - name: tenant-operator
   repository: file://../../operators/deploy/tenant-operator
   version: 0.1.0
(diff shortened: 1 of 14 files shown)


frisso (approved):

frisso:
Just to know, does this require some changes to the documentation as well?
E.g., something here? https://github.com/netgroup-polito/CrownLabs/tree/master/infrastructure/identity-provider#accessing-k8s-cluster-using-keycloak-as-authentication-server

giorio94:
> Just to know, does this require some changes to the documentation as well?
> E.g., something here? https://github.com/netgroup-polito/CrownLabs/tree/master/infrastructure/identity-provider#accessing-k8s-cluster-using-keycloak-as-authentication-server

No, this is related only to the authentication in front of the user instances, and it is completely transparent from the users' point of view. Overall, this change is required in view of the authentication for the new dashboard. Still, I was wondering whether to take the chance to completely change this approach, to reduce the number of object created for each new VM and save resources in the cluster.

frisso:
For me, any improvement is welcome.
The only limit, as usual, is the manpower required to achieve the above objective, given the high amount of work that we're already carrying on.

QcFe:
Does it make sense to test this one again?

giorio94:
> Does it make sense to test this one again?

Definitely

QcFe:
I fear there's some issue with `websiteBaseUrl` instead of `websiteBaseUri` and the deployment of the operator

giorio94:
> I fear there's some issue with `websiteBaseUrl` instead of `websiteBaseUri` and the deployment of the operator

That was expected, since the inventories in CrownOps are still old. I changed that URI to URL for consistency (though I probably named that so). And the new flag is not configured at all.

GabriFila (approved):

PR #496, opened 2021-07-08
Infrastructure: add oauth2-proxy installation docs

giorio94:
# Description

This PR adds the documentation regarding the installation of the centralized oauth2-proxy deployment, which is used to prevent access to the graphical desktop of the user instances to unauthenticated users. Related to #490 

# How Has This Been Tested?

- [x] Performing the deployment of oath2-proxy in the cluster

### DIFF
--- a/infrastructure/identity-provider/README.md
+++ b/infrastructure/identity-provider/README.md
@@ -280,3 +280,35 @@ users:
       command: kubectl
       env: null
 ```
+
+## User Instances Authentication
+
+In CrownLabs, the access to the graphical desktop of the user instances should be protected, so that only authenticated users can connect to them.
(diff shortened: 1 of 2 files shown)


QcFe (approved):
Looks good to me. Was just thinking if it's the case "generalize" the crownlabs domains with the usual example.com

giorio94:
> Looks good to me. Was just thinking if it's the case "generalize" the crownlabs domains with the usual example.com

That is something we have never been really that consistent throughout the repo. Still, as for the infrastructural manifests we usually kept the correct domain, to make them as similar as possible to the actual ones. Would keep so for the moment, since the URLs are anyhow public in the end. Differently, I personally prefer the *fake* ones inside the Helm charts, to make it more evident we are using the values from CrownOps.

GabriFila (approved):

QcFe:
\rebase

PR #499, opened 2021-07-08
Add operators parallelism to helm charts and fix logs

QcFe:
# Description
This PR adds support to set the number of concurrent reconciles for the instance and tenant operators through helm chart values.
Moreover the tenant operator helm chart has been cleaned up to remove the configmap, in order to match the deployment of the instance operator (the secret containing nextcloud and keycloak login data remains). 
This PR also includes a little fix by @giorio94 to prevent the instance operator from make unnecessarily verbose logs in some circumstances (whose fault is also mine).


GabriFila (approved):

giorio94 (approved):
/lgtm 

You should just add the new parameters also to the general values file

PR #507, opened 2021-07-14
Fix for GraphQL server

Ralls0:
# Description

Some fixes for security and checking of the token.

### DIFF
--- a/qlkube/src/watch.js
+++ b/qlkube/src/watch.js
@@ -1,6 +1,7 @@
 const k8s = require('@kubernetes/client-node');
 const fetch = require('node-fetch');
-const { publishEvent } = require('./pubsub.js');
+const { publishEvent } = require('./pubsub');
+const { graphqlLogger } = require('./utils');
 
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
@@ -15,6 +16,7 @@ async function canWatchResource(
   namespace,
   name
 ) {
+  graphqlLogger(`[i] Check resource ${resource} in namespace ${namespace}`);
   return fetch(
     `${apiServerUrl}/apis/authorization.k8s.io/v1/selfsubjectaccessreviews`,
     {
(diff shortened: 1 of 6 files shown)


qlkube/src/watch.js
```diff
@@ -69,7 +69,11 @@ function kwatch(api, label) {
         });
       },
       err => {
-        console.error(err);
+        if (err.message === 'Not Found') {
+          setTimeout(() => kwatch(api, label), 2000);
+        } else {
+          console.error(err);
```
GabriFila, line 75: Add a `console.error("No items found to watch, restarting in 2s")` before the `setTimeout`
And replace the `console.error(err)` with a  `console.error("Error when watching",err)`

qlkube/src/apolloServerPlugin.js
```diff
@@ -0,0 +1,40 @@
+const requestStartRegistry = {
+  // Fires whenever a GraphQL request is received from a client.
+  async requestDidStart(requestContext) {
+    if (requestContext.request.operationName !== 'IntrospectionQuery')
+      console.log(
+        `Request started!
```
GabriFila, line 6: I would add the `\n` for better readability

qlkube/src/apolloServerPlugin.js
```diff
@@ -0,0 +1,40 @@
+const requestStartRegistry = {
```
GabriFila, line 1: Use a more descriptive name

giorio94 (approved):
AFAIU, /lgtm

qlkube/src/watch.js
```diff
@@ -63,13 +65,19 @@ function kwatch(api, label) {
...
       (type, apiObj, watchObj) => {
+        graphqlLogger(`[i] Publish event on ${label} label`);
         publishEvent(label, {
           apiObj,
           type,
         });
       },
       err => {
-        console.error(err);
+        if (err.message === 'Not Found') {
```
QcFe, line 75: Using error texts to distinguish error types might not be a great deal. Try to see fi there's another field inside which can be more "robust". If this is an http response error there should be a `code` property or something alike, which in this case should be 404. Otherwise if it's a special class (like, SomethingNotFoundError) you could do `if (err instanceof SomethingNotFoundError) ...`

QcFe (approved):
Just a couple of minor things but looks alright

PR #510, opened 2021-07-15
Added Navbar component

cheina97:
Components added:

- Navbar (common)

### DIFF
--- a/frontend/src/Decorators.tsx
+++ b/frontend/src/Decorators.tsx
@@ -5,6 +5,8 @@ import { Layout, Skeleton } from 'antd';
 import Button from 'antd-button-color';
 import { StoryContext } from '@storybook/react/dist/ts3.9/client/preview/types';
 import { Story } from '@storybook/react';
+import { MemoryRouter, Route } from 'react-router-dom';
+import ThemeContextProvider from './contexts/ThemeContext';
 
 const ModalDecorator = (Story: Story, context: StoryContext) => {
   const [show, setShow] = useState(false);
@@ -39,4 +41,23 @@ const LogoDecorator = (Story: Story) => (
   </div>
 );
 
-export { ModalDecorator, LogoDecorator };
+const NavbarDecorator = (Story: Story) => (
+  <MemoryRouter>
+    <ThemeContextProvider>
+      <Route path="/">
+        <Layout>
+          <Story />
+          <Layout.Content>
+            <div className="m-4">
+              <Skeleton />
+              <Skeleton />
+              <Skeleton />
+            </div>
+          </Layout.Content>
+        </Layout>
+      </Route>
+    </ThemeContextProvider>
+  </MemoryRouter>
+);
+
+export { ModalDecorator, NavbarDecorator, LogoDecorator };
--- a/frontend/src/components/common/Navbar/Navbar.tsx
+++ b/frontend/src/components/common/Navbar/Navbar.tsx
@@ -1,0 +1,166 @@
+import { FC, useContext } from 'react';
+import { Layout, Drawer } from 'antd';
+import Button from 'antd-button-color';
+import { MenuOutlined } from '@ant-design/icons';
+import { Link, useLocation } from 'react-router-dom';
+import { useState } from 'react';
+import ThemeSwitcher from '../../misc/ThemeSwitcher';
+import { ThemeContext } from '../../../contexts/ThemeContext';
+import './Navbar.less';
+import Logo from '../Logo';
+import { LogoutButton } from '../LogoutButton';
+
+const Header = Layout.Header;
+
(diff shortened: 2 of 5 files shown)


QcFe:
I'm not really convinced with this approach actually. To keep things decoupled I'd avoid working with history here at all. I'd rather pass a function from the parent with the setter of the current view. Then such parent will be responsible of handling such change. 
(Also I'd make small custom components and/or adding a loop to avoid that heavy code replication)
Any thought @GabriFila?

cheina97:
> I'm not really convinced with this approach actually. To keep things decoupled I'd avoid working with history here at all. I'd rather pass a function from the parent with the setter of the current view. Then such parent will be responsible of handling such change.
> (Also I'd make small custom components and/or adding a loop to avoid that heavy code replication)
> Any thought @GabriFila?

Ok i'll find a way to avoid code duplication

frontend/src/components/common/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const Navbar: FC<INavbarProps> = ({ ...props }) => {
+  const { activeLabel, setActiveLabel } = props;
+  const [show, setShow] = useState(false);
+
+  const buttons = [
+    { path: '/', name: 'main' },
+    { path: '/active', name: 'active' },
+    { path: '/drive', name: 'drive' },
+    { path: '/account', name: 'account' },
```
QcFe, line 24: I'd actually put the display label here, so with the uppercase already, then pass and use setActiveRoute from the parent instead of setActiveLabel

QcFe (dismissed):

frontend/src/components/common/Navbar/Navbar.tsx
cheina97: But how can i change the passed state clicking the buttons without a setState function? Maybe i'm misunderstanding

cheina97:
Reopen PR

frontend/src/components/common/Logo/Logo.tsx
```diff
@@ -0,0 +1,23 @@
...
+export interface ILogoProps {
+  width: number;
+}
+
+const Logo: FC<ILogoProps> = ({ ...props }) => {
+  const { isDarkTheme } = useContext(ThemeContext);
+  const { width } = props;
+  return (
+    <img
+      src={isDarkTheme ? logoDarkTheme : logoLightTheme}
```
giorio94, line 16: Since the difference between the two logos seems to be only related to color, could it be changed dynamically through CSS instead of having two versions?

frontend/src/components/common/Logo/Logo.tsx
```diff
@@ -0,0 +1,23 @@
...
+      width={width}
+      alt="Crownlabs Logo"
```
giorio94, line 18:
```suggestion
      alt="CrownLabs Logo"
```

frontend/src/Decorators.tsx
```diff
@@ -1,8 +1,12 @@
+/* eslint-disable react/no-multi-comp */
```
giorio94, line 1: Should add a comment why this is needed.

giorio94 (approved):
From a user perspective, looks mostly good to me.

Was just wondering whether to put the theme switcher inside the hamburger menu in the mobile view. WDYT?

frontend/src/Decorators.tsx
cheina97: I just added a comment. After this PR, as agreed with gabriele, we will do a "refactoring" of the decorators by dividing them into several files. This file will be removed.

frontend/src/components/common/Logo/Logo.tsx
cheina97: I tried to change it using CSS, but without success. Maybe would be better to discuss about it during the meeting. According to the react documentation https://it.reactjs.org/docs/context.html#when-to-use-context is a good idea to use a context to choose the used theme. Maybe i could be try to hardcode the svg in the logo component to avoid file duplication (i'm talking about logo.svg and logoLightTheme.svg)

cheina97:
> From a user perspective, looks mostly good to me.
> 
> Was just wondering whether to put the theme switcher inside the hamburger menu in the mobile view. WDYT?

I added an alternative behaviour for the responsive mode. I modified and pushed app.tsx just to show you the correct behaviour of the component, that doesn't work correctly in the storybook. Before merging this PR i will restore the old app.tsx

giorio94:
I personally like the current implementation. Do you think an intermediate status with only main, active and the hamburger for the others could be useful? I was mainly thinking about a split screen scenario, with the navbar which now looks empty.

cheina97:
deploy-staging

PR #514, opened 2021-07-22
Keycloak version upgrade

QcFe:
# Description

This PR upgrades keycloak deployment documentation in order to reflect the one currently deployed in the CrownLabs cluster.


infrastructure/identity-provider/conf-files/keycloak-configuration.yaml
```diff
@@ -1,325 +1,555 @@
...
+  podAntiAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      - labelSelector:
+          matchLabels:
+            {{- include "keycloak.selectorLabels" . | nindent 10 }}
+          matchExpressions:
+            - key: app.kubernetes.io/component
+              operator: NotIn
+              values:
+                - test
```
giorio94, line 152: Should be configured with the correct label, otherwise it is useless. The `preferredDuringSchedulingIgnoredDuringExecution` term below should be commented, since we have only one zone

giorio94 (commented):
Seems mostly fine to me. In addition to the inline comment, please update also the documentation to use the standard installation command (i.e., without pulling the helm repo, but referencing the value file), and drop the reference to the ingress creation.

giorio94 (approved):

frisso (approved):

PR #519, opened 2021-07-28
Instanceset landing: dynamic instance creation and redirection

QcFe:
# Description

This PR introduces the concept of the InstanceSet. This will primarily be used for managing exams (and exams demos, actually called "exercise") through CrownLabs.
The InstanceSet consists in a hardenizable (through network policies and other possible security related precautions) namespace (deployed by a possible InstanceSet operator) which initially includes at least the deployment of the landing redirector included in this PR. Such component is responsible for verifying the correctness of the request (by checking compliancy to the polito's didattica portal identifiers generation) then possibly creating the instance on demand (if specified by the `--dynamic` switch) and eventually redirecting to the generated instance's endpoint.

# How Has This Been Tested?
- [x] Locally
- [x] On cluster


operators/pkg/instanceset-landing/security.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func generateEndpoint(identifier string) string {
+	identifier = strings.ToLower(identifier)
+
+	// Get result and encode as hexadecimal string
+	return fmt.Sprintf("https://%ss.%s/%s/",
+		Options.Type,
+		Options.BaseHostName,
+		Options.CourseCode+"/"+keyedShaFromString(identifier),
```
giorio94, line 70: Should not concatenate here but as part of the printf

operators/pkg/instanceset-landing/handler.go
```diff
@@ -0,0 +1,61 @@
...
+		w.WriteHeader(400)
+		fmt.Fprint(w, "Invalid request")
+		klog.Errorf("Invalid request: invalid username [%v]", usernameParam)
+		return
+	}
+
+	examEndpoint := generateEndpoint(username)
+
+	if Options.DynamicStartup {
+		var err error
```
giorio94, line 42:
```suggestion
```

operators/pkg/instanceset-landing/handler.go
```diff
@@ -0,0 +1,61 @@
...
+		examEndpoint, err = enforceOnDemandInstance(username)
+		if err != nil {
+			w.WriteHeader(500)
+			fmt.Fprintf(w, "OnDemand instance creation failed: %v", err)
```
giorio94, line 46:
```suggestion
			fmt.Fprintf(w, "OnDemand instance creation/retrieval failed")
```

Should not propagate internal error conditions to the client.

operators/pkg/instanceset-landing/creation.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// DatasourcePvcClaimName is the name of the pvc containing the initial content of the instance.
+const DatasourcePvcClaimName = "crownlabs-exam-datasource"
+
+func generateInstanceName(matricola, courseID string) string {
+	return "ex-" + courseID + "-" + matricola + "-crownlabs"
+}
+
+func enforceOnDemandInstance(identifier string) (string, error) {
```
giorio94, line 20: Would try to be consistent. Either this is always identifier or usename.

operators/pkg/instanceset-landing/creation.go
```diff
@@ -0,0 +1,64 @@
...
+	matricola, courseID, err := decodeIdentifier(identifier)
+
+	if err != nil {
+		klog.Errorf("Couldn't verify identifier [%v]", err)
+		return "", errors.New("invalid_identifier")
```
giorio94, line 25: Would propagate the original error

operators/pkg/instanceset-landing/creation.go
```diff
@@ -0,0 +1,64 @@
...
+	instanceName := generateInstanceName(matricola, courseID)
+	klog.V(3).Infof("- Creating instance %s for user %s", instanceName, matricola)
+
+	// TODO: handle dataSource if Options.StartupOpts.InitialContentSourceURL
+
+	instance := &clv1alpha1.Instance{}
+	instance.SetName(instanceName)
+	instance.SetNamespace(Options.Namespace)
+
+	_, err = ctrl.CreateOrUpdate(context.Background(), Client, instance, func() error {
```
giorio94, line 37: 1. Should use the request context, and not a forged one
2. Should not discard the result, to print an appropriate log message

operators/pkg/instanceset-landing/creation.go
```diff
@@ -0,0 +1,64 @@
...
+		if instance.CreationTimestamp.IsZero() {
+			instance.Spec = clv1alpha1.InstanceSpec{
+				Template: clv1alpha1.GenericRef{
```
giorio94, line 40: Should be set to running

operators/pkg/instanceset-landing/security.go
```diff
@@ -0,0 +1,72 @@
...
+func keyedShaFromString(str string) string {
+	h := hmac.New(sha256.New, []byte(Options.InternalKey))
+
+	_, err := h.Write([]byte("SEC-" + str + "-CL"))
+	if err != nil {
+		klog.Error("SHA256 generation error [%v]", err)
+	}
+
+	return fmt.Sprintf("%x", h.Sum(nil))
+}
```
giorio94, line 25: Can be dropped ATM

operators/pkg/instanceset-landing/security.go
```diff
@@ -0,0 +1,72 @@
...
+
+func sha1FromString(in string) string {
+	h := sha1.New() // nolint:gosec // used in polito api
+	_, err := h.Write([]byte(in))
+	if err != nil {
+		klog.Error("SHA256 generation error [%v]", err)
```
giorio94, line 31:
```suggestion
		klog.Error("SHA1 generation error [%v]", err)
```

operators/pkg/instanceset-landing/security.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return fmt.Sprintf("%x", h.Sum(nil))
+}
+
+func generateIstanceIdentifier(matr, cCode string) string {
```
giorio94, line 37: Please, be consistent with the parameters names. Would also change the method to checkIdentifierValidity, or something like that.

giorio94 (commented):
Overall, this looks mostly OK to me (comments are really minor/stylistic). In general, I would try to hide all the interaction with the PoliTO APIs in a single file, though an appropriate interface (e.g. with three methods such as `ValidateQuery`, `GenerateEndpoint` and `GenerateInstanceName`), so that is could be easily generalized.

operators/pkg/instanceset-landing/handler.go
QcFe: this is here to create a new examEndpoint below, is there any better way to handle this?

giorio94: Got it. I thought you could just define it below together with the initialization, but I missed you would also shadow at the same time the other `examEndpoint` variable.

QcFe: Found another way

frisso:
Given that a new resource has been created, I would suggest to update the documentation section with a brief description of this object.

giorio94:
> Given that a new resource has been created, I would suggest to update the documentation section with a brief description of this object.

Currently, the description of this PR is a bit misleading, since a new resource has not yet been added (but left as a future step). INO, the documentation should be added in that PR.

operators/pkg/instanceset-landing/options.go
```diff
@@ -0,0 +1,63 @@
...
+var Options options
+
+// Initialize flags and associate each parameter to the given options object.
+func (o *options) Init() {
+	flag.StringVar(&o.ListenerAddr, "address", ":8080", "[address]:port of the landing server")
+	flag.StringVar(&o.CourseCode, "course", "", "Course code")
+	flag.StringVar(&o.Template.Name, "template-name", "", "CrownLabs Template for the exam")
+	flag.StringVar(&o.Template.Namespace, "template-namespace", "workspace-exams", "Namespace of CrownLabs Template for the exam")
+	flag.StringVar(&o.Namespace, "namespace", "", "Custom namespace name, if not given it's automatically generated")
+	flag.StringVar(&o.HostName, "hostname", "crownlabs.polito.it", "CrownLabs instanceset host")
```
giorio94, line 33: Is this still required or could it be dropped?

giorio94 (approved):
/lgtm

Just a couple of minor comments

operators/pkg/instanceset-landing/options.go
QcFe: IIRC that was to give instances a custom hostname so that it can be selectively unlocked on LDB. It can be removed now and put back later if you think it's better

giorio94: Since we were planning to add that info to the template it would no longer be needed,

PR #522, opened 2021-07-31
Adding keycloak auth to frontend 

GabriFila:
# Description

This PR adds the necessary elements for the keycloak authentication to work

# How Has This Been Tested?

This has been tested locally, waiting for the staging environment


frontend/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useState,
+} from 'react';
+import {
+  REACT_APP_CROWNLABS_OIDC_PROVIDER_URL,
+  REACT_APP_CROWNLABS_OIDC_CLIENT_ID,
+} from '../env';
+interface IAuthContext {
+  isLoggedIn: undefined | boolean;
+  token: undefined | string;
+  userId: string | undefined;
```
QcFe, line 16: Isn't something like this working? 
```suggestion
  isLoggedIn?: boolean;
  token?: string;
  userId?: string;
```
If not, couldn't we still use empty strings to say those values aren't present?
(also do we need isLoggedIn to be undefined?)

QcFe (approved):
It looks good to me, just a minor question

frontend/src/contexts/AuthContext.tsx
GabriFila: We use `undefined` for 2 reasons:
- with the isLoggedin, at first you don't actually know if the user is logged in or not, so I used it there, to hva more control. Although, in the case of our app, the user can't do anything if not logged in, so it could be a simple boolean, where the false means "unknown or not logged in"
- for the other cases I prefer to have undefined instead of an empty string to have a more explicit code. And yes, I think we can use that syntax

frontend/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+          setIsLoggedIn(false);
+          setToken(undefined);
+          setUserId(undefined);
+        }
+        kc.loadUserInfo()
+          .then((res: any) => {
+            setUserId(res.preferred_username);
+          })
+          .catch(err => {
+            console.error('ERROR WHEN GETTING USER INFO', err);
```
giorio94, line 54: Nit. Could be de-capitalized.

frontend/public/config.js
```diff
@@ -7,5 +7,3 @@
 window.REACT_APP_CROWNLABS_GRAPHQL_URL = undefined;
 window.REACT_APP_CROWNLABS_OIDC_PROVIDER_URL = undefined;
 window.REACT_APP_CROWNLABS_OIDC_CLIENT_ID = undefined;
-window.REACT_APP_CROWNLABS_OIDC_CLIENT_SECRET = undefined;
-window.REACT_APP_CROWNLABS_OIDC_REDIRECT_URI = undefined;
```
giorio94, line 11: Should also be removed from the helm templates.

frontend/public/index.html
```diff
@@ -25,7 +25,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Crownlabs</title>
+    <title>CrownLabs</title>
```
giorio94, line 28: Could you please update also the meta description above?

frontend/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export const AuthContext = createContext<IAuthContext>({
+  isLoggedIn: false,
+  token: undefined,
+  userId: undefined,
+});
+
+const kc = Keycloak({
+  url: REACT_APP_CROWNLABS_OIDC_PROVIDER_URL,
+  realm: 'crownlabs',
```
giorio94, line 27: Could be added as environment variable to avoid the hardcoded value

giorio94 (approved):
/lgtm 

Just a couple of minor comments inline. Don't know whether it is worth to investigate about the authentication refresh in this PR or in another one.

PR #523, opened 2021-08-02
Fix for wrapped fields

Ralls0:
# Description

This PR contains some fixing for wrapped fields into the subscription. So, now the wrapped fields can be watched and a unique subscription object is returned.

### DIFF
--- a/qlkube/deploy/qlkube/values.yaml
+++ b/qlkube/deploy/qlkube/values.yaml
@@ -18,22 +18,35 @@ configuration:
   exposedAPIs:
     fileName: apiGroups.js
     apis:
-    - crownlabs.polito.it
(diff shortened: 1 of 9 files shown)


qlkube/deploy/qlkube/values.yaml
```diff
@@ -8,7 +8,7 @@ image:
   repository: crownlabs/qlkube
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: ""
+  tag: ''
```
giorio94, line 11: Could you please revert the quote changes.

giorio94 (approved):
/lgtm

Just a couple of nits inline.

PR #532, opened 2021-08-08
Decorators and storybook refactoring

cheina97:
This PR delete the Decorators.tsx file ( to avoid conflicts during merges and rebases ), and split its content in more files. It also updates component stories in order to exploit the refactored decorators.

Now decorators become more generics and their name describe their behaviour to avoid duplicated ones. However in some cases specific decorators for specific components still have to be used.

This PR also modify old storybooks hiding some useless controls.


frontend/src/decorators/ClickAndShowDecorator.tsx
```diff
@@ -0,0 +1,36 @@
...
+          >
+            Open Modal
+          </Button>
+        </div>
+      </Layout.Content>
+    </Layout>
+  );
+};
+
+export { ClickAndShowDecorator };
```
GabriFila, line 36: I would call this `DialogOpenDecorator`

GabriFila (approved):

PR #533, opened 2021-08-09
Instance operator: add initial ready time metric

giorio94:
# Description

This PR introduces a prometheus metric to analyze the time required by the instances to become ready upon creation. The same value is also shown as an additional field of the instance itself.

# How Has This Been Tested?

- [x] Staging environment

### DIFF
--- a/operators/pkg/instance-controller/metrics.go
+++ b/operators/pkg/instance-controller/metrics.go
@@ -1,0 +1,50 @@
+// Copyright 2020-2021 Politecnico di Torino
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package instance_controller
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	metricInitialReadyTimesLabelWorkspace   = "workspace"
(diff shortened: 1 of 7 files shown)


operators/pkg/instance-controller/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+			metricInitialReadyTimesLabelWorkspace,
+			metricInitialReadyTimesLabelTemplate,
+			metricInitialReadyTimesLabelEnvironment,
+			metricInitialReadyTimesLabelType,
+			metricInitialReadyTimesLabelPersistent,
+		},
+	)
+)
+
+func init() {
```
QcFe, line 46: I fear I'm missing where this method is called from

operators/pkg/instance-controller/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+	metricInitialReadyTimesLabelEnvironment = "environment"
+	metricInitialReadyTimesLabelType        = "type"
+	metricInitialReadyTimesLabelPersistent  = "persistent"
+)
+
+var (
+	metricInitialReadyTimes = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "instance_initial_ready_time_second",
+		Help:    "The number of seconds required by the Instances to become ready upon creation",
+		Buckets: append(prometheus.LinearBuckets(5, 5, 11), prometheus.LinearBuckets(60, 30, 19)...),
```
QcFe, line 34: Is there any documentation in repo for what this line does? Could it be useful to add any reference to it as a comment for this line?

QcFe (commented):
I'm missing a couple of points

operators/pkg/instance-controller/metrics.go
giorio94: The `init` method is special, since it is executed automatically when the package is imported. https://golang.org/doc/effective_go#init

operators/pkg/instance-controller/metrics.go
giorio94: Check if this is what you meant.

QcFe: Ah totally makes sense now

QcFe (approved):

PR #541, opened 2021-08-16
Updated Logo component: added fixedColor mode

cheina97:
Components updated:

- Logo (common)

This PR adds the fixedColor prop, which allows you to set a fixed color for the logo, which will not change according to the selected theme.
This feature can be useful in CornerButton component, where button color don't change according to the  current theme

### DIFF
--- a/frontend/src/components/common/Logo/Logo.tsx
+++ b/frontend/src/components/common/Logo/Logo.tsx
@@ -5,17 +5,17 @@ import { ReactComponent as SvgLogo } from '../../../assets/logo.svg';
 export interface ILogoProps {
   widthPx?: number;
   className?: string;
+  color?: string;
 }
 
 const Logo: FC<ILogoProps> = ({ ...props }) => {
-  const { widthPx, className } = props;
+  const { widthPx, className, color } = props;
   return (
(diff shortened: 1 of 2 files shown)


QcFe (approved):
looks good, I would simply put `color` instead of `fixedColor` though

frontend/src/components/common/Logo/Logo.tsx
```diff
@@ -5,15 +5,17 @@ import { ReactComponent as SvgLogo } from '../../../assets/logo.svg';
...
   widthPx?: number;
   className?: string;
+  color?: string;
 }
 
 const Logo: FC<ILogoProps> = ({ ...props }) => {
-  const { widthPx, className } = props;
+  const { widthPx, className, color } = props;
   return (
     <div className={'flex items-center justify-center m-0 p-0 ' + className}>
```
QcFe, line 14: Are we sure we want to keep the logo with this div wrapper? (with such flex stuff too?)

QcFe (approved):
For me this modification is alright, not sure about the div though

frontend/src/components/common/Logo/Logo.tsx
cheina97: I removed it. It seems work correctly without it.

PR #551, opened 2021-08-27
Updated Navbar: bg-color and shadow

cheina97:
Components updated:

- Navbar (common)

This PR changes the Navbar componen's background color in light mode and add a shadow under the navbar.

### DIFF
--- a/frontend/src/components/common/Navbar/Navbar.less
+++ b/frontend/src/components/common/Navbar/Navbar.less
@@ -11,7 +11,7 @@
 }
 
 body.light .navbar-bg {
-  background-color: #f2f2f2;
+  background-color: white;
 }
 
 .navbar-bg-transparent {
@@ -19,7 +19,7 @@ body.light .navbar-bg {
 }
 
 body.light .navbar-bg-transparent {
-  background-color: white;
+  background-color: #f0f2f5;
 }
 
 .navbar-button {
@@ -46,4 +46,12 @@ body.light .navbar-button:focus {
 
 body.light .navbar-title {
   color: @button-light-hover;
+}
+
+.cl-navbar{
+  --cl-bg-navbar: #141414;
+
+  body.light &{
+    --cl-bg-navbar: white;
+  }
 }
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/Navbar/Navbar.tsx
```diff
@@ -137,7 +137,7 @@ const Navbar: FC<INavbarProps> = ({ ...props }) => {
       <Drawer
         className={' block ' + (routes.length > 4 ? 'lg:hidden' : 'md:hidden')}
         bodyStyle={{
-          backgroundColor: isDarkTheme ? '#141414' : '#f2f2f2',
+          backgroundColor: isDarkTheme ? '#141414' : '#ffffff',
```
QcFe, line 140: Can't this be done in pure css? We know the teme based on the root class name

frontend/src/components/common/Navbar/Navbar.less
```diff
@@ -47,3 +47,11 @@ body.light .navbar-button:focus {
...
   color: @button-light-hover;
 }
+
+.cl-navbar {
+  background-color: #141414;
+
+  body.light {
+    background-color: white;
+  }
+}
```
QcFe, line 57: This approach seems much more effective to me. However I'm not convinced it could work, shouldn't it be like this?
```suggestion
.cl-navbar {
  background-color: #141414;
}
body.light .cl-navbar {
    background-color: white;
  }

```

cannarelladev: your suggestion is right even if don't change the final result. Outside this suggestion we've noted this change (from hard-coded style to dynamic css classes) produces some problems to the "trasparence effect " of the drawer cause ANT has a very strange behavior. So we'd know if it's a good idea to apply this change breaking the right behavior of the drawer. Unfortunately we think that there isn't a good solution for everything: we've to decide if accept a bad hard code or a good code with some problems

QcFe (approved):

PR #552, opened 2021-08-27
Added TemplatesTableLogic component

cheina97:
Components added:

- TemplatesTableLogic

This PR add the TemplatesTableLogic component, whose purpose is to get the informations about  workspaces templates and pass them to the TemplatesTable component.

### DIFF
--- a/frontend/src/components/workspaces/DashboardLogic/DashboardLogic.tsx
+++ b/frontend/src/components/workspaces/DashboardLogic/DashboardLogic.tsx
@@ -15,16 +15,23 @@ function DashboardLogic() {
   //startPolling(20000);
 
   return !loading && data && !error ? (
-    <Dashboard
-      workspaces={
-        data.tenant?.spec?.workspaces?.map(workspace => {
-          return {
-            workspaceId: workspace?.workspaceRef?.workspaceId as string,
-            role: workspace?.role!,
-          };
-        }) ?? []
-      }
-    />
+    <>
+      <Dashboard
+        workspaces={
+          data.tenant?.spec?.workspaces?.map(workspace => {
+            return {
+              workspaceId: workspace?.workspaceRef?.workspaceWrapper
+                ?.itPolitoCrownlabsV1alpha1Workspace?.spec
+                ?.workspaceName as string,
+              role: workspace?.role!,
+              workspaceNamespace: workspace?.workspaceRef?.workspaceWrapper
+                ?.itPolitoCrownlabsV1alpha1Workspace?.status?.namespace
+                ?.workspaceNamespace!,
+            };
+          }) ?? []
+        }
+      />
+    </>
   ) : (
     <div className="h-full w-full flex justify-center items-center">
       <Spin size="large" />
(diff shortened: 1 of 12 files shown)


frontend/src/components/workspaces/DashboardLogic/DashboardLogic.tsx
```diff
@@ -15,16 +15,23 @@ function DashboardLogic() {
...
-    />
+    <>
+      <Dashboard
+        workspaces={
+          data.tenant?.spec?.workspaces?.map(workspace => {
+            return {
+              workspaceId: workspace?.workspaceRef?.workspaceWrapper
+                ?.itPolitoCrownlabsV1alpha1Workspace?.spec
+                ?.workspaceName as string,
+              role: workspace?.role!,
```
GabriFila, line 26: Why use the `?` and the `!` to access the role

GabriFila, line 26: Why use the `?` and the `!` to access the role

frontend/src/components/workspaces/Templates/TemplatesTableRow/TemplatesTableRow.tsx
```diff
@@ -41,15 +41,9 @@ const TemplatesTableRow: FC<ITemplatesTableRowProps> = ({ ...props }) => {
       <div className="w-full flex justify-between py-0">
         <Space size={'middle'}>
           {gui ? (
-            <DesktopOutlined
-              className={'primary-color-fg'}
-              style={{ fontSize: '24px' }}
-            />
+            <DesktopOutlined style={{ fontSize: '24px', color: '#1c7afd' }} />
```
GabriFila, line 44: Why did you hardcoded the values?

frontend/src/components/workspaces/Templates/TemplatesTableLogic/TemplatesTableLogic.tsx
```diff
@@ -0,0 +1,118 @@
...
+    loading: loadingInstances,
+    error: errorInstances,
+    startPolling: startPollingInstances,
+  } = useOwnedInstancesQuery({
+    variables: { tenantNamespace: `tenant-${userId}` },
+  });
+
+  startPollingInstances(500);
+
+  useEffect(() => {
```
GabriFila, line 46: If this is not temporary (maybe you are waiting for the graphql watchers)
Add some comments to this file to explain all the logic regarding `reload`, `refetchTemplate`, etc

frontend/src/components/workspaces/DashboardLogic/DashboardLogic.tsx
cheina97: The `?` is because the value returned by the query could be null, the `!` is because we trust that the value won't be null.

frontend/src/components/workspaces/Templates/TemplatesTableRow/TemplatesTableRow.tsx
cheina97: It's not the first time we hardcode the `fontSize` css property. We have also hardcoded the `color` property before, in the `WorkspaceGridItem`. We did it for the other icons so we decided to do it also for this one. If u want i can open a PR to refactor all icons.

frontend/src/components/workspaces/DashboardLogic/DashboardLogic.tsx
GabriFila: Oh i thought that the second one would make the first one useless, sorry

frontend/src/components/workspaces/Templates/TemplatesTableRow/TemplatesTableRow.tsx
GabriFila: I think we'll need to do it, but now now

GabriFila (approved):

PR #586, opened 2021-09-18
Added Instances subscriptions

cheina97:
Components modified:

- TemplatesTableLogic

This pr add the subscription to the query used to retrieve data about instances and generate a notification when an instance goes in "VmiReady" phase.


QcFe (approved):
/lgtm

PR #589, opened 2021-09-19
Style fixes

cheina97:
Modified components:

- Navbar
- TemplatesTableRow
- AppLayout

This PR add a tooltip with info to the logout button and the theme switcher. 
It also change the persistent icon used in TemplatesTableRow with an infinite symbol and change the info button icon.


QcFe (approved):
Looks alright. A minor note: the Internet seems more prone to the use of the word _disservice_ rather than _misservice_ but don't really know if it makes a difference.

Issue #512, opened 2021-07-19
Instance operator refactoring


giorio94:
## Overview

This issue summarizes the main tasks towards the instance operator refactoring, for increased robustness and modularity.

## Task list

* [x] Rationalize the usage of the `CreateOrUpdate` function;
* [x] Replace the deprecated `setInstanceStatus` with a single update at the end of the reconcile;
* [X] Replace the custom VM readiness check with a proper readiness probe;
* [X] Move the `DataVolume` definition to the VM spec, instead of a separated object;
* [x] Move all resource definitions to the `forge` package, split in smaller units for easier reuse;
* [x] Migrate to structured logging;
* [x] Increment the test coverage;
* [x] Standardise file/dirs/package names;

### `CreateOrUpdate` guidelines

As for the  `CreateOrUpdate` usage, it is necessary to balance between resource enforcement and user disruption. For this reason, it is suggested the following:
* Do not modify those specs (e.g., deployment/VM templates) which would cause the restart of the user environment;
* Do enforce the various metadata information (labels, annotations, references), while ensuring user values are conserved;
* Vary the approach in the other cases depending on the specific situation, trying to reduce disruption in case the logic changed after a new release. Document whatever approach is selected for future information.

giorio94:
Closing, as fully addressed

PR #593, opened 2021-09-22
Instance controller reconcile tests refactor

QcFe:
# Description

This PR refactors the instance controller tests in order to use a simpler environment (use a simple kubernetes client without cache, by not using controller manager) that avoids the need to wait for the cache to propagate (which required the doesEventuallyExist function) and removes possible conflict which occasionally made tests fail.

Ref. #512


operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
+				Email: "test@email.me",
 			},
-			Spec:   v1.NamespaceSpec{},
-			Status: v1.NamespaceStatus{},
 		}
-		webdavSecret = v1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "webdav-secret",
-				Namespace: InstanceNamespace,
+		workspace = clv1alpha1.Workspace{ObjectMeta: metav1.ObjectMeta{Name: testName}}
```
giorio94, line 96: Would avoid creating if not used

operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
-					Name:      TemplateName + "persistent",
-					Namespace: TemplateNamespace,
-				},
-				Tenant: crownlabsv1alpha2.GenericRef{
-					Name: TenantName,
-				},
-			},
-			Status: crownlabsv1alpha2.InstanceStatus{},
+
+		Expect(k8sClient.Create(ctx, &ns)).ToNot(HaveOccurred())
```
giorio94, line 113:
```suggestion
		Expect(k8sClient.Create(ctx, &ns)).To(Succeed())
```

Nit. Just a bit more idiomatic (same for all other occurrences)

operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
-		TemplateNamespace = "template-namespace"
-		TenantName        = "test-tenant"
+	RunReconciler := func() error {
+		_, err := instanceReconciler.Reconcile(ctx, reconcile.Request{
+			NamespacedName: forge.NamespacedName(&instance),
+		})
+		if err != nil {
+			return err
+		}
+		Expect(err).NotTo(HaveOccurred())
```
giorio94, line 67:
```suggestion
```

operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
+// The following are integration tests aiming to verify that the instance controller reconcile method correctly creates the various resources it has to manage, by running several combinations of configuration. This is not exaustive as a more granular coverage is achieved through the unit tests in this package.
+var _ = Describe("The instance-controller Reconcile method", func() {
+	ctx := context.Background()
+	var (
+		testName        string
+		runInstance     bool
+		ns              corev1.Namespace
+		tenant          clv1alpha1.Tenant
+		workspace       clv1alpha1.Workspace
+		template        clv1alpha2.Template
```
giorio94, line 49: Would avoid defining here the ones never accesses after before/creation

operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
+				testName = "test-container-persistent"
+				environment = clv1alpha2.Environment{
+					Name:            testName,
+					Image:           "docker-image:v0",
+					EnvironmentType: clv1alpha2.ClassContainer,
+					GuiEnabled:      true,
+					Resources:       testResources,
+					Mode:            clv1alpha2.ModeStandard,
+					Persistent:      true,
 				}
```
giorio94, line 139: Nit. Could move a level up to reduce duplication, and just overwrite the necessary fields.

operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
+				By("Asserting the right exposition resources exist", func() {
+					Expect(k8sClient.Get(ctx, forge.NamespacedName(&instance), &service)).NotTo(HaveOccurred())
+					Expect(k8sClient.Get(ctx, forge.NamespacedNameWithSuffix(&instance, forge.IngressGUINameSuffix), &ingress)).NotTo(HaveOccurred())
+					Expect(k8sClient.Get(ctx, forge.NamespacedNameWithSuffix(&instance, forge.IngressMyDriveNameSuffix), &ingress)).To(BeNotFoundError())
+				})
+
+				By("Asserting the state is coherent", func() {
+					// Check the status phase is unset since it's retrieved from the VM (and the kubervirt operator is not available in the test env)
+					Expect(instance.Status.Phase).To(Equal(clv1alpha2.EnvironmentPhaseUnset))
+				})
```
giorio94, line 299: Could do the same as below to force it to running

operators/pkg/instance-controller/controller_test.go
```diff
@@ -12,290 +12,408 @@
...
+	}
+}
+
+func (s *IsNotFoundErrorMatcher) FailureMessage(actual interface{}) (message string) {
+	return format.Message(actual, "to match NotFoundError")
+}
+
+func (s *IsNotFoundErrorMatcher) NegatedFailureMessage(actual interface{}) (message string) {
+	return format.Message(actual, "not to match NotFoundError")
 }
```
giorio94, line 419: Thumbs-up for this. Would just move to an utility package, to make it easier to be reused

giorio94 (approved):
/lgtm

Just some minor comments inline

PR #599, opened 2021-10-05
Add apply mutation 

Ralls0:
# Description

Add _apply mutation_ and some fields into the query like `creationTimestamp`

### DIFF
--- a/frontend/codegen.yml
+++ b/frontend/codegen.yml
@@ -1,5 +1,6 @@
 overwrite: true
-schema: "https://graphql.preprod.crownlabs.polito.it/"
+# the schema needs to be manually downloaded.
+schema: schema.graphql
 documents: './src/**/*.{graphql,ts}'
 generates:
   src/generated-types.tsx:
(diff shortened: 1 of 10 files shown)


frontend/src/graphql-components/mutation/editTemplate.mutation.graphql
```diff
@@ -0,0 +1,30 @@
+mutation editTemplate(
```
giorio94, line 1: For consistency would either call this applyTemplate, or the others edit*

frontend/codegen.yml
```diff
@@ -1,5 +1,5 @@
 overwrite: true
-schema: "https://graphql.preprod.crownlabs.polito.it/"
+schema: schema.graphql
```
giorio94, line 2: Is this change intentional? If so, would add a comment about the fact that the schema needs to be manually downloaded.

giorio94 (approved):
/lgtm

PR #641, opened 2021-11-11
Add manager parameter, and name and namespace where were missing

Ralls0:
# Description

Add manager parameter, and name and namespace where were missing


QcFe:
Saw the new queries, however I'd need all namespace fields not to be renamed, so it should be called always namespace in every query...

QcFe (approved):

QcFe (approved):

PR #648, opened 2021-11-15
SSH bastion: support spaces in ssh keys' comments

luco5826:
# Description

This PR enables the bastion to undestand keys that have comments with spaces inside. 

Without this PR, keys with spaces in the comment were unable to be used to connect to VMs through the bastion

# How Has This Been Tested?

`go test` returns a successful status code


giorio94 (approved):
/lgtm

PR #668, opened 2021-11-24
Add checkLabelSelector

Ralls0:
# Description

Add checkLabelSelector

### DIFF
--- a/qlkube/src/decorateSubscription.js
+++ b/qlkube/src/decorateSubscription.js
@@ -28,9 +28,73 @@ function checkMetadata(metadataItemA, metadataItemB) {
   );
 }
 
-function checkLabelSelector(subVar, payloadVar) {
-  // TODO: check on LabelSelector
-  return true;
+function splitLabelSelector(labelSelector) {
+  let key;
+  let values = [];
+
+  /*
+   * Split labelSelector into key and values
+   * First case: label selector into the form "crownlabs.polito.it/workspace=sid"
+   * Second case: label selector into the form "crownlabs.polito.it/workspace in (sid)"
+   */
+  if (labelSelector.includes('=')) {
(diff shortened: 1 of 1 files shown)


qlkube/src/decorateSubscription.js
```diff
@@ -28,9 +28,79 @@ function checkMetadata(metadataItemA, metadataItemB) {
...
+   * The resource has no labels, so no match
+   */
+  if (!labels) {
+    graphqlLogger(
+      `[i] (${ruid}) The result of the check is false due to empty labels`
+    );
+    return false;
+  }
+
+  if (typeof labels === 'string') labels = JSON.parse(labels);
```
QcFe, line 89: It's better to wrap this in a try/catch block, I guess the labels variable could contain forged contents and stuff alike. `JSON.parse` throws an exception if malformed JSON input is read. If you have an external try/catch you don't need to put one here, otherwise the whole process might be killed.

qlkube/src/decorateSubscription.js
```diff
@@ -28,9 +28,79 @@ function checkMetadata(metadataItemA, metadataItemB) {
...
+
+  const { key, values } = splitLabelSelector(labelSelector);
+
+  for (const k in labels) {
+    if (k.includes(key)) {
+      values.forEach(v => {
+        result = result || v === labels[k];
+      });
```
QcFe, line 97: You could probably do something like
```js
result |= values.includes(labels[k]);
```

QcFe (approved):
Minor style bits

qlkube/src/decorateSubscription.js
Ralls0: To be honest, the `labels` variable is already an object, this is a forgetfulness. I remove this line.

QcFe (approved):

PR #670, opened 2021-11-26
Minor fixes

cheina97:
Components updated:

- Navbar
- TemplatesTable

This PR unify the arrow icon of the tables and prevent theNavbarMenu component from beeing rendered if the tenant query hasn' finished or if there are errors.

### DIFF
--- a/frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
+++ b/frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
@@ -2,7 +2,7 @@
 import { FC, useEffect } from 'react';
 import { Table } from 'antd';
 import { TemplatesTableRow } from '../TemplatesTableRow';
-import { RightOutlined } from '@ant-design/icons';
+import { CaretRightOutlined } from '@ant-design/icons';
 import { useState } from 'react';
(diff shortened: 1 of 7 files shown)


frontend/src/components/workspaces/Templates/TemplatesTable/TemplatesTable.tsx
```diff
@@ -107,7 +107,7 @@ const TemplatesTable: FC<ITemplatesTableProps> = ({ ...props }) => {
           // eslint-disable-next-line react/no-multi-comp
           expandIcon: ({ expanded, onExpand, record }) =>
             record.instances.length ? (
-              <RightOutlined
+              <CaretRightOutlined
```
QcFe, line 110:
```suggestion
              <CaretRightOutlined className="transition-transform"
```

cheina97: Sorry i forgot to answer you. I haven't added that class because it doesn't seem to work

QcFe: We could have a broken Tailwind install then I guess...
I think we could manually add [this code](https://tailwindcss.com/docs/transition-property) to our codebase.

QcFe (approved):

PR #676, opened 2021-12-02
Provisioning: minor modifications to the Kubernetes ansible role

giorio94:
# Description

This PR updates the Kubernetes version in the ansible role, and restores the default MTU in Docker.


frisso:
Looks perfect to me.
I would simply add in the description of this PR why the MTU was changed to 1440, and now restored to 1500.

frisso (approved):

PR #671, opened 2021-11-27
Change flags for disabling webhooks, Keycloak, andNC integrations

CosimoMichelagnoli:
### Description

This pr simplifies  local tests of tenant-operator by disabling  nextcloud, keycloak and webhook. Done by managing values passed through run-tenant command in Makefile (e.g. --kc-url=DISABLE) 

Co-authored by: Grazia D'Onghia graziadonghia925@gmail.com

PR #681, opened 2021-12-02
Fix another bug introduced in #671

giorio94:
# Description

This PR fixes a bug introduced in #671, which caused the keycloak client to be never initialized.


QcFe (approved):
😳

PR #687, opened 2021-12-06
Clock fix

cannarelladev:
# Description

Just a minor fix on the CL App Clock, during instance creation there was usually a visual glitch that showed a negative AGE


giorio94:
Just a minor note. It is not guaranteed client and server timers are synchronized. Would make sure negative values are shown as `now` to prevent glitches in case of clock skews.

cannarelladev:
@giorio94 dome

cannarelladev:
> Just a minor note. It is not guaranteed client and server timers are synchronized. Would make sure negative values are shown as `now` to prevent glitches in case of clock skews.

@giorio94 Done!

PR #697, opened 2021-12-09
Template Table Logic: templateId fix

cheina97:
This PR changes the field used to distinguish each template. Now the unique id of a template is built concatenating the workspace name and the template name.

### DIFF
--- a/frontend/src/graphql-components/query/instances.query.graphql
+++ b/frontend/src/graphql-components/query/instances.query.graphql
@@ -1,3 +1,4 @@
+# Write your query or mutation here
 query ownedInstances($tenantNamespace: String!) {
   instanceList: listCrownlabsPolitoItV1alpha2NamespacedInstance(
     namespace: $tenantNamespace
@@ -7,6 +8,7 @@ query ownedInstances($tenantNamespace: String!) {
         name
         namespace
         creationTimestamp
+        labels
       }
       status {
         ip
--- a/frontend/src/utilsLogic.tsx
+++ b/frontend/src/utilsLogic.tsx
@@ -7,6 +7,7 @@ import {
   Exact,
   ItPolitoCrownlabsV1alpha2Instance,
   ItPolitoCrownlabsV1alpha2Template,
+  Maybe,
   OwnedInstancesQuery,
   UpdatedOwnedInstancesSubscriptionResult,
   UpdatedWorkspaceTemplatesSubscriptionResult,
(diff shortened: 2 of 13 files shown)


frontend/src/utilsLogic.tsx
```diff
@@ -292,6 +301,10 @@ export const getWorkspacesMapped = (
     .filter(ws => ws.templates.length);
 };
 
+export const makeTemplateKey = (tid: string, wid: string) => {
+  return `${tid}-${wid}`;
+};
+
```
QcFe, line 307:
```suggestion
export const makeTemplateKey = (tid: string, wid: string) => `${tid}-${wid}`;

```

frontend/src/utilsLogic.tsx
```diff
@@ -216,7 +223,8 @@ export const getManagerInstances = (
     prettyName: spec?.prettyName,
     gui: guiEnabled,
     persistent: persistent,
-    idTemplate: name,
+    templateName: name,
+    templateId: makeTemplateKey(name, namespace.replace(/^workspace-/, '')),
```
QcFe, line 227: It should be better to get the workspace name from the labels. If this isn't possible it can be left like this...

frontend/src/graphql-components/query/instances.query.graphql
```diff
@@ -7,6 +8,7 @@ query ownedInstances($tenantNamespace: String!) {
         name
         namespace
         creationTimestamp
+        labels
```
QcFe, line 11:
```suggestion
        labels {
          workspaceName
        }
```
Something alike could actually even work (clearly with the right key, I don't remember which it is now), would you mind trying it out?

cheina97: I tried and it doesn't work

Ralls0: `labels` field contains a `JSON` object, not a graphQL type. So, you should parse the given JSON and work on it.

cannarelladev: but, could be better adding to the query the same fields about workspace that are in Instances Label selector query? so that the response should be a graphql type and not a JSON

Ralls0: Pay attention, valid label keys have two segments: an optional prefix and name, separated by a slash `/` (see [labels and selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)). So, in this case, checking labels with kubectl you should see something like that: `"crownlabs.polito.it/workspace: sid"`, but the output given by GraphQL will be something like this `"crownlabsPolitoItWorkspace": "sid"`.  I don't know the reason, but `.` and `/` are removed and the words are written in camelCase.

cheina97: I solved it, it returns an object and it's not necessary to parse it, now I'm casting it with a function to an explicit type.

cheina97: Another problem was that I didn't add the label field to the subscriptions. Now it seems to be ok

frontend/src/utilsLogic.tsx
```diff
@@ -192,7 +213,9 @@ export const joinInstancesAndTemplates = (
 ) =>
   templates.map(t => ({
     ...t,
-    instances: instances.filter(i => i.idTemplate === t.id),
+    instances: instances.filter(
+      i => i.templateId === makeTemplateKey(t.id, t.workspaceId!)
```
QcFe, line 217: Also here, it's better to accept _undefinable_ parameters and ensuring they're undefined.

frontend/src/utilsLogic.tsx
```diff
@@ -218,8 +241,12 @@ export const getManagerInstances = (
     prettyName: spec?.prettyName,
     gui: guiEnabled,
     persistent: persistent,
-    idTemplate: templateName,
     templatePrettyName: prettyName,
+    templateName: templateName,
+    templateId: makeTemplateKey(
+      templateName,
+      templateNamespace.replace(/^workspace-/, '')
```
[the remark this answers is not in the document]
cheina97: Ask to @cannarelladev

frontend/src/utilsLogic.tsx
cheina97: I solved it avoiding undefined value in makeTemplateKey function

frontend/src/utilsLogic.tsx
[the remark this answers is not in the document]
QcFe: Please then add a TODO on this line for this

QcFe (approved):

PR #700, opened 2021-12-15
ErrorsQueue avoid duplicated errors

cheina97:
This PR adds a control in the **dispatchError** function to prevent the same error to be added more times.

### DIFF
--- a/frontend/src/components/debug/ThrowExceptionPage/ThrowExceptionPage.tsx
+++ b/frontend/src/components/debug/ThrowExceptionPage/ThrowExceptionPage.tsx
@@ -1,21 +1,52 @@
 import { Button } from 'antd';
-import { FC, useState } from 'react';
+import { FC, useContext } from 'react';
+import { ErrorContext } from '../../../errorHandling/ErrorContext';
+import { ErrorTypes } from '../../../errorHandling/utils';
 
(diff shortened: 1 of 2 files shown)


frontend/src/errorHandling/RenderErrorAlert/RenderErrorAlert.tsx
```diff
@@ -2,22 +2,26 @@ import { FC, useContext } from 'react';
...
       <div className="flex w-full justify-center">
         <Result
           className="p-0"
           //It's not a real 500 error, that's just for visualize a cute image with the component
           status="500"
-          title="Application error."
+          title={`Application error, you have ${
+            errorsQueue.filter(e => e.getType() === ErrorTypes.RenderError)
+              .length
+          } errors in queue`}
```
QcFe, line 24: Would leave it as it was

QcFe (commented):
Clean it up and we should be there

frontend/src/errorHandling/RenderErrorAlert/RenderErrorAlert.tsx
cheina97: Oki

QcFe (approved):

PR #703, opened 2021-12-16
CrownLabs Exams Agent

QcFe:
# Description

This PR adds the logic for the new CrownLabs Exam Agent, necessary for the integration with Moodle.


.github/workflows/build.yml
```diff
@@ -79,6 +79,9 @@ jobs:
           # InstanceSet tools
           - instanceset-landing
 
+          # Exams agent
+          - exmagnt
```
giorio94, line 83: Vowels do not cost more than consonants ;)

operators/cmd/exmagnt/main.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/klog/v2"
+
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/exmagnt"
+)
+
+func main() {
+	exmagnt.Options.Init()
+	exmagnt.Options.Parse()
+
+	if err := exmagnt.Options.Validate(); err != nil {
```
giorio94, line 31: Could merge `Parse` and `Validate` in one call only

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+
+	clv1alpha2 "github.com/netgroup-polito/CrownLabs/operators/api/v1alpha2"
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/forge"
+)
+
+//go:embed redirecting.html
+var httpPageStartingUp string
+
+// EAInstance represents an Instance within the exmagnt.
+type EAInstance struct {
```
giorio94, line 44: EA is already implied by the package name, and does not need to be repeated. The same for all other occurrences.

operators/pkg/exmagnt/options.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Perform general flags validation.
+func (o *options) Validate() error {
+	if o.TemplatesNS == "" {
+		return errors.New("missing argument: templates-namespace")
+	}
+
+	if o.AllowedIPs == "" {
+		klog.Warningln("No whitelist IPs have been specified: all IPs are allowed")
```
giorio94, line 65: Would say this is an info message, rather than a warning

operators/pkg/exmagnt/options.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Options object holds all the instanceset parameters.
+var Options options
+
+// Initialize flags and associate each parameter to the given options object.
+func (o *options) Init() {
+	flag.StringVar(&o.ListenerAddr, "address", ":8888", "[address]:port of the landing server")
+	flag.StringVar(&o.TemplatesNS, "templates-namespace", "", "Namespace of CrownLabs Templates that will be listed")
+	flag.StringVar(&o.AllowedIPs, "allowed-ips", "", "Comma separated list of IPs that are allowed to create new instances")
```
giorio94, line 44: Could probably leverage something like this to automatically perform IP validation: https://github.com/liqotech/liqo/blob/master/pkg/utils/args/cidrlist.go

operators/pkg/exmagnt/options.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return nil
+}
+
+// AllowedIPs contains a list of IPs that are allowed to create new instances.
+type AllowedIPs string
+
+// CheckAllowedIP checks if the given IP is allowed within the AllowedIPs.
+func (o *options) CheckAllowedIP(r *http.Request) bool {
```
giorio94, line 79: Would take directly the IP as parameters, to decouple this from the rest of the implementation.

operators/pkg/exmagnt/options.go
```diff
@@ -0,0 +1,91 @@
...
+	if o.AllowedIPs == "" {
+		return true
+	}
+
+	for _, allowedIP := range strings.Split(o.AllowedIPs, ",") {
+		if glob.Glob(allowedIP, r.RemoteAddr) {
```
giorio94, line 85: Would use CIDRs instead of globs in this context

operators/pkg/exmagnt/template.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	// EATemplateEndpoint is the endpoint for the EAInstance.
+	EATemplateEndpoint = "/template"
+)
+
+// EATemplate represents a Template within the exmagnt.
+type EATemplate struct {
+	Name       string `json:"name"`
+	PrettyName string `json:"prettyName"`
+	CreatedAt  string `json:"createdAt"`
```
giorio94, line 41: You can probably leave this as timestamp, and then it is formatted automatically.

operators/pkg/exmagnt/template.go
```diff
@@ -0,0 +1,76 @@
...
+	if r.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		fmt.Fprintf(w, "Method not allowed")
+		return
+	}
+
+	templates := clv1alpha2.TemplateList{}
+	if err := Client.List(r.Context(), &templates, client.InNamespace(Options.TemplatesNS)); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		klog.Error(err)
```
giorio94, line 55: Could be worth enrich a bit the error message

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+	ID                string                               `json:"id"`
+	Template          string                               `json:"template"`
+	Running           *bool                                `json:"running,omitempty"`
+	CustomizationUrls clv1alpha2.InstanceCustomizationUrls `json:"customizationUrls"`
+}
+
+// EAInstanceHandler is the handler for the EAInstance.
+type EAInstanceHandler struct{}
```
giorio94, line 52: Would save here the client, instead of using the singleton. It simplifies future extensions.

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprint(w, "Bad request")
+		return
+	}
+
+	inst := EmptyInstanceFromRequest(r)
+	instSpec := InstanceSpecFromEAInstance(&eaInstance)
+
+	op, err := ctrl.CreateOrUpdate(r.Context(), Client, inst, func() error {
+		inst.Spec = instSpec
```
giorio94, line 140:
```suggestion
		inst.Spec = InstanceSpecFromEAInstance(&eaInstance)
```

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+	})
+
+	if err != nil {
+		klog.Errorf("Instance %s cannot be %s: %v", inst.Name, op, err)
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "Instance %s cannot be %s", inst.Name, op)
+		return
+	}
+
+	w.Header().Add("Content-Type", "text/plain")
```
giorio94, line 151: It is probably better to always stick to json (BTW, might it be worth returning the created instance information?)

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+
+	switch op {
+	case ctrlutil.OperationResultCreated:
+		w.WriteHeader(http.StatusCreated)
+	case ctrlutil.OperationResultUpdated:
+		w.WriteHeader(http.StatusOK)
+	default:
+		w.WriteHeader(http.StatusInternalServerError)
```
giorio94, line 159: Why is this a server error? It could be simply unmodified.

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+	Running           *bool                                `json:"running,omitempty"`
+	CustomizationUrls clv1alpha2.InstanceCustomizationUrls `json:"customizationUrls"`
+}
+
+// EAInstanceHandler is the handler for the EAInstance.
+type EAInstanceHandler struct{}
+
+const (
+	// EAInstanceEndpoint is the endpoint for the EAInstance.
+	EAInstanceEndpoint = "/instance"
```
giorio94, line 56: I personally feel this should be plural, but I suppose it would then need to be changed also on the other side.

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	fmt.Fprint(w, op)
+	klog.Errorf("Instance %s has been %s", inst.Name, op)
+}
+
+// HandleGetAll handles the GET request for all the instances.
+func HandleGetAll(w http.ResponseWriter, r *http.Request) {
+	var instances clv1alpha2.InstanceList
+	if err := Client.List(r.Context(), &instances); err != nil {
```
giorio94, line 169: Shouldn't filter only the current namespace?

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+		klog.Errorf("Error encoding instances: %v", err)
+		return
+	}
+}
+
+// GetInstanceIDFromRequest returns the instance id from the request.
+func GetInstanceIDFromRequest(r *http.Request) string {
+	InstanceEP := path.Join(Options.BasePath, EAInstanceEndpoint) + "/"
+	instID := strings.Replace(r.URL.Path, InstanceEP, "", 1)
+	return instID
```
giorio94, line 194: Hoped there was a better way

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+			Name: clv1alpha2.SVCTenantName,
+		},
+		PrettyName:        fmt.Sprintf("Exam %s", instReq.ID),
+		CustomizationUrls: instReq.CustomizationUrls,
+	}
+}
+
+// EAInstanceFromInstance creates an EAInstance from a given Instance.
+func EAInstanceFromInstance(inst *clv1alpha2.Instance) *EAInstance {
+	instSpec := inst.Spec
```
giorio94, line 263: Seems useless to me

operators/pkg/exmagnt/instance.go
```diff
@@ -0,0 +1,270 @@
...
+	err := json.NewDecoder(r.Body).Decode(&inst)
+	return inst, err
+}
+
+// InstanceSpecFromEAInstance creates an InstanceSpec from a given EAInstance.
+func InstanceSpecFromEAInstance(instReq *EAInstance) clv1alpha2.InstanceSpec {
+	running := true
+	if instReq.Running != nil {
+		running = *instReq.Running
+	}
```
giorio94, line 246: You can use `pointer.BoolDeref(..., true))`

giorio94 (commented):
Some comments inline. The logic seems fine to me though

operators/pkg/exmagnt/options.go
QcFe: a eh actually made it myself before seeing this link...

operators/pkg/exmagnt/instance.go
QcFe: it's used in the get all and now also as the put response

operators/pkg/exmagnt/instance.go
QcFe: can think of any?

operators/pkg/exmagnt/instance.go
QcFe: we could expose both, then we remove the singular after the change has been made on the other side

operators/cmd/examagent/main.go
```diff
@@ -0,0 +1,67 @@
...
+	logger := klogr.NewWithOptions().WithName("main")
+
+	if err := examagent.Options.Parse(); err != nil {
+		logger.Error(err, "invalid configuration")
+		os.Exit(1)
+	}
+
+	k8sClient, err := examagent.NewK8sClient()
+	if err != nil {
+		logger.Error(err, "unable to prepare k8s client")
```
giorio94, line 40: Should also exit

operators/cmd/examagent/main.go
```diff
@@ -0,0 +1,67 @@
...
+	)
+
+	http.HandleFunc("/healthz", healthzHandler)
+
+	http.Handle(InstanceEP, examagent.InstanceHandler{Log: logger.WithName("instance-handler"), Client: k8sClient})
+
+	http.Handle(TemplateEP, examagent.TemplateHandler{Log: logger.WithName("template-handler"), Client: k8sClient})
+
+	klog.Info("Instanceset landing listening on port ", examagent.Options.ListenerAddr)
+	klog.Fatal(http.ListenAndServe(examagent.Options.ListenerAddr, nil))
```
giorio94, line 55: Still to be updated

operators/pkg/examagent/instance.go
```diff
@@ -0,0 +1,289 @@
...
+const (
+	// InstanceAdapterEndpoint is the endpoint for the InstanceAdapter.
+	InstanceAdapterEndpoint = "/instance"
+
+	// XForwardedFor -> "X-Forwarded-For" header.
+	XForwardedFor = "X-Forwarded-For"
+)
+
+// ServeHTTP is the Instance handler for the examagent.
+func (ih InstanceHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
giorio94, line 67:
```suggestion
func (ih *InstanceHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```

I suppose. In alternative, is this is requested so, I would change the others for consistency.

operators/pkg/examagent/instance.go
```diff
@@ -0,0 +1,289 @@
...
+	log := ih.Log.WithValues("remote-ip", r.Header.Get(XForwardedFor), "method", r.Method)
```
giorio94, line 68: Might it be worth to also add the path as value?

operators/pkg/examagent/instance.go
```diff
@@ -0,0 +1,289 @@
...
+	log.Info("sending starting-up page")
+
+	w.Header().Add("refresh", "5")
+	w.WriteHeader(http.StatusCreated)
+	fmt.Fprint(w, httpPageStartingUp)
+}
+
+// HandlePut handles the PUT request for a InstanceAdapter api call.
+func (ih *InstanceHandler) HandlePut(w http.ResponseWriter, r *http.Request, log logr.Logger) {
+	if err := Options.CheckAllowedIP(r.Header.Get("X-Forwarded-For")); err != nil {
```
giorio94, line 132: Should use the constant

operators/pkg/examagent/instance.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// EmptyInstanceFromRequest returns an Instance from a given request with just the ObjectMeta field set.
+func EmptyInstanceFromRequest(r *http.Request) *clv1alpha2.Instance {
+	instID := GetInstanceIDFromRequest(r)
+
+	inst := &clv1alpha2.Instance{
+		ObjectMeta: metav1.ObjectMeta{Name: instID, Namespace: Options.TemplatesNS},
+	}
+	return inst
```
giorio94, line 246:
```suggestion
	return &clv1alpha2.Instance{
		ObjectMeta: metav1.ObjectMeta{Name: GetInstanceIDFromRequest(r), Namespace: Options.TemplatesNS},
	}
```

Nit. Just a bit shorter

operators/pkg/examagent/instance.go
```diff
@@ -0,0 +1,289 @@
...
+			Name: clv1alpha2.SVCTenantName,
+		},
+		PrettyName:        fmt.Sprintf("Exam %s", instReq.ID),
+		CustomizationUrls: &instReq.CustomizationUrls,
+	}
+}
+
+// AdapterFromInstance creates an InstanceAdapter from a given Instance.
+func AdapterFromInstance(inst *clv1alpha2.Instance) *InstanceAdapter {
+	instSpec := inst.Spec
```
giorio94, line 282: Is this alias necessary?

operators/pkg/examagent/options.go
```diff
@@ -0,0 +1,104 @@
...
+
+// CheckAllowedIP checks if the given IP is allowed within the AllowedIPs.
+func (o *options) CheckAllowedIP(rawIP string) error {
+	if o.AllowedIPs == "" {
+		return nil
+	}
+
+	addr := net.ParseIP(rawIP)
+	if addr == nil {
+		return fmt.Errorf("invalid X-Forwarded-For header (%v)", rawIP)
```
giorio94, line 94:
```suggestion
		return fmt.Errorf("invalid IP (%v)", rawIP)
```

This is agnostic from how it was retrieved

operators/pkg/examagent/template.go
```diff
@@ -0,0 +1,83 @@
...
+// TemplateAdapter represents a Template within the examagent.
+type TemplateAdapter struct {
+	Name       string      `json:"name"`
+	PrettyName string      `json:"prettyName"`
+	CreatedAt  metav1.Time `json:"createdAt"`
+}
+
+// ServeHTTP is the handler for the TemplateAdapter.
+func (th TemplateHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	log := th.Log.WithValues("remote-ip", r.Header.Get(XForwardedFor))
```
giorio94, line 49: Would also add the path and the method

operators/pkg/examagent/instance.go
```diff
@@ -0,0 +1,289 @@
...
+		return
+	}
+}
+
+// GetInstanceIDFromRequest returns the instance id from the request.
+func GetInstanceIDFromRequest(r *http.Request) string {
+	InstanceEP := path.Join(Options.BasePath, InstanceAdapterEndpoint) + "/"
+	instID := strings.Replace(r.URL.Path, InstanceEP, "", 1)
+	return instID
+}
```
giorio94, line 216: Would move this after the delete method, to segregate receiver and non-receiver methods.

operators/pkg/examagent/options.go
```diff
@@ -0,0 +1,104 @@
...
+	ipNets           []*net.IPNet
+}
+
+// Options object holds all the instanceset parameters.
+var Options options
+
+// Initialize flags and associate each parameter to the given options object.
+func (o *options) Init() {
+	flag.StringVar(&o.ListenerAddr, "address", ":8888", "[address]:port of the landing server")
+	flag.StringVar(&o.TemplatesNS, "templates-namespace", "", "Namespace of CrownLabs Templates that will be listed")
```
giorio94, line 44: This is probably just namespace, since everything is inside that (both instances and templates)

giorio94 (commented):
A few more comments

operators/pkg/exmagnt/instance.go
giorio94: I was just referring to that line, not the entire function.. Sorry

operators/pkg/exmagnt/options.go
giorio94: The outcome is slightly different, as with the other approach it is the library itself that returns the error during the flag parsing operation (printing also the appropriate message), instead of doing that in a post processing step. Still it is also fine as you did for this use-case

operators/pkg/examagent/instance.go
QcFe: It is required to be this way, still using the pointer where it's possible isn't it better in terms of performance? I don't really know how big are the objects inside...

operators/pkg/exmagnt/instance.go
giorio94: None actually. Would just save InstanceEP as field of the handler structure, to avoid having it hardcoded here (wouldn't allow to support two paths for instance)

giorio94 (approved):

PR #713, opened 2021-12-24
Added operator-selector label to the sample tenant namespace

cheina97:
This PR adds the **crownlabs.polito.it/operator-selector** label in the tenant namespaces generated by the example files in samples, in order to allow the instance operator to start in a local cluster (e.g. kind, minikube...  )

### DIFF
--- a/operators/samples/instance.yaml
+++ b/operators/samples/instance.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: tenant-john-doe
+  labels:
+    crownlabs.polito.it/operator-selector: local
 ---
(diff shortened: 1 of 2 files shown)


operators/samples/instance.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: tenant-john-doe
+  labels:
+    crownlabs.polito.it/operator-selector: local
```
giorio94, line 6: At the moment, this is not required, as the check is performed on the namespace and not the instance itself

giorio94 (approved):

PR #714, opened 2021-12-24
Added Selective group destroy

cannarelladev:
# Description

This PR adds the possibility to select a group of instances to be destroyed

### DIFF
--- a/frontend/src/components/activePage/ActiveView/ActiveView.tsx
+++ b/frontend/src/components/activePage/ActiveView/ActiveView.tsx
@@ -1,25 +1,19 @@
+import { DeleteOutlined } from '@ant-design/icons';
+import { Col, Space } from 'antd';
+import Button from 'antd-button-color';
 import { FC, useEffect, useState } from 'react';
-import { Space, Col, Input, Tooltip, Checkbox, Popover } from 'antd';
-import ViewModeButton from './ViewModeButton/ViewModeButton';
-import Box from '../../common/Box';
 import { User, WorkspaceRole } from '../../../utils';
+import { SessionValue, StorageKeys } from '../../../utilsStorage';
+import Box from '../../common/Box';
+import { ModalAlert } from '../../common/ModalAlert';
 import TableInstanceLogic from '../TableInstance/TableInstanceLogic';
 import TableWorkspaceLogic from '../TableWorkspaceLogic/TableWorkspaceLogic';
-import Button from 'antd-button-color';
-import {
-  CloseOutlined,
-  DeleteOutlined,
-  FullscreenExitOutlined,
-  FullscreenOutlined,
-  SearchOutlined,
-} from '@ant-design/icons';
-import { SessionValue, StorageKeys } from '../../../utilsStorage';
-import { ModalAlert } from '../../common/ModalAlert';
+import Toolbox from '../Toolbox/Toolbox';
+import ViewModeButton from './ViewModeButton/ViewModeButton';
 
 const view = new SessionValue(StorageKeys.Active_View, WorkspaceRole.user);
 const advanced = new SessionValue(StorageKeys.Active_Headers, 'true');
 
-const { Search } = Input;
 export interface IActiveViewProps {
   user: User;
   workspaces: Array<{
@@ -38,7 +32,6 @@ const ActiveView: FC<IActiveViewProps> = ({ ...props }) => {
   const [destroySelectedTrigger, setDestroySelectedTrigger] = useState(false);
   const [showAlert, setShowAlert] = useState(false);
   const [searchField, setSearchField] = useState('');
-  const [searchPopover, setSearchPopover] = useState(false);
   const [currentView, setCurrentView] = useState<WorkspaceRole>(
     managerView ? (view.get() as WorkspaceRole) : WorkspaceRole.user
   );
@@ -46,13 +39,25 @@ const ActiveView: FC<IActiveViewProps> = ({ ...props }) => {
     !managerView || advanced.get() !== 'false'
   );
   const [showCheckbox, setShowCheckbox] = useState(false);
-
   const [selectiveDestroy, setSelectiveDestroy] = useState<string[]>([]);
 
   const selectToDestroy = (instanceId: string) => {
     selectiveDestroy.includes(instanceId)
       ? setSelectiveDestroy(old => old.filter(id => id !== instanceId))
       : setSelectiveDestroy(old => [...old, instanceId]);
+  };
+
+  const deselectAll = () => setSelectiveDestroy([]);
+
+  const displayCheckbox = () => {
+    if (!showCheckbox) {
+      setShowCheckbox(true);
+    } else {
+      setShowCheckbox(() => {
+        deselectAll();
+        return false;
+      });
+    }
   };
 
   useEffect(() => {
@@ -78,7 +83,10 @@ const ActiveView: FC<IActiveViewProps> = ({ ...props }) => {
             size="middle"
             icon={<DeleteOutlined />}
             className="border-0"
-            onClick={() => setDestroySelectedTrigger(true)}
+            onClick={() => {
+              setDestroySelectedTrigger(true);
+              setShowAlert(false);
+            }}
           >
             Destroy Selected
           </Button>,
@@ -94,77 +102,7 @@ const ActiveView: FC<IActiveViewProps> = ({ ...props }) => {
               </p>
             </div>
           ) : (
-            managerView &&
-            currentView === WorkspaceRole.manager && (
-              <div className="h-full flex justify-center items-center gap-4 px-5">
-                <Button
-                  className="hidden xl:block"
-                  type="primary"
-                  shape="round"
-                  size="middle"
-                  icon={<FullscreenOutlined />}
-                  onClick={() => setExpandAll(true)}
-                >
-                  Expand
-                </Button>
-                <Button
-                  className="hidden xl:block"
-                  type="primary"
-                  shape="round"
-                  size="middle"
-                  icon={<FullscreenExitOutlined />}
-                  onClick={() => setCollapseAll(true)}
-                >
-                  Collapse
-                </Button>
-                <Tooltip title="Expand All" placement="top">
-                  <Button
-                    className="xl:hidden"
-                    type="primary"
-                    shape="circle"
-                    size="middle"
-                    icon={<FullscreenOutlined />}
-                    onClick={() => setExpandAll(true)}
-                  />
-                </Tooltip>
-                <Tooltip title="Collapse All" placement="top">
-                  <Button
-                    className="xl:hidden"
-                    type="dark"
-                    shape="circle"
-                    size="middle"
-                    icon={<FullscreenExitOutlined />}
-                    onClick={() => setCollapseAll(true)}
-                  />
-                </Tooltip>
-                <Checkbox
-                  checked={showAdvanced}
-                  onChange={e => setShowAdvanced(e.target.checked)}
-                >
-                  <strong>Show Headers</strong>
-                </Checkbox>
-                <Checkbox
-                  checked={showCheckbox}
-                  onChange={e => setShowCheckbox(e.target.checked)}
-                >
-                  <strong>Show Checkbox</strong>
-                </Checkbox>
-                <Button
-                  className="hidden xl:block"
-                  type="primary"
-                  ghost
-                  shape="round"
-                  size="middle"
-                  icon={<DeleteOutlined />}
-                  onClick={e => {
-                    e.stopPropagation();
-                    setShowAlert(true);
-                  }}
-                >
-                  Destroy Selected{` (${selectiveDestroy.length})`}
-                </Button>
-              </div>
-            )
+            ''
           ),
           size: 'middle',
           right: managerView && (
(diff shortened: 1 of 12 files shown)


frontend/src/components/activePage/TableInstance/TableInstance.tsx
```diff
@@ -58,6 +64,27 @@ const TableInstance: FC<ITableInstanceProps> = ({ ...props }) => {
...
+
+  const selectGroup = () => {
+    if (checked) {
+      if (indeterminate) {
+        const tmp = instances.filter(i => !selectiveDestroy?.includes(i.id));
+        tmp.forEach(i => selectToDestroy!(i.id));
+      } else {
+        const tmp = instances.filter(i => selectiveDestroy?.includes(i.id));
+        tmp.forEach(i => selectToDestroy!(i.id));
+      }
```
QcFe, line 82:
```suggestion
      instances.filter(i => indeterminate ^ selectiveDestroy?.includes(i.id)).forEach(i => selectToDestroy!(i.id));
```
Could be cleaner if this actually worked...

cannarelladev:
Ready for review

frontend/src/components/activePage/TableInstance/RowInstanceHeader/RowInstanceHeader.tsx
```diff
@@ -94,7 +102,21 @@ const RowInstanceHeader: FC<IRowInstanceHeaderProps> = ({ ...props }) => {
...
       >
         <div className="w-full flex justify-start items-center">
-          <div className="flex items-center justify-center w-16">
+          {showCheckbox && (
+            <div className={`flex items-center justify-center w-12`}>
+              <Checkbox
+                checked={checked}
+                className="p-0"
+                indeterminate={indeterminate && checked}
+                onClick={() => selectGroup()}
```
QcFe, line 111:
```suggestion
                onClick={selectGroup}
```

frontend/src/components/activePage/ActiveView/ActiveView.tsx
```diff
@@ -33,14 +29,38 @@ const ActiveView: FC<IActiveViewProps> = ({ ...props }) => {
...
+
+  const selectToDestroy = (instanceId: string) => {
+    selectiveDestroy.includes(instanceId)
+      ? setSelectiveDestroy(old => old.filter(id => id !== instanceId))
+      : setSelectiveDestroy(old => [...old, instanceId]);
+  };
+
+  const deselectAll = () => {
+    setSelectiveDestroy([]);
+  };
```
QcFe, line 52:
```suggestion
  const deselectAll = () => setSelectiveDestroy([]);
```

frontend/src/components/activePage/TableInstance/TableInstance.tsx
```diff
@@ -58,6 +64,30 @@ const TableInstance: FC<ITableInstanceProps> = ({ ...props }) => {
     });
   };
 
+  const checked = !!instances.filter(i => selectiveDestroy?.includes(i.id))
```
QcFe, line 67: Would extract `instances.filter(i => selectiveDestroy?.includes(i.id))` into a variable since it's used more than once and it's quite verbose...

frontend/src/components/activePage/TableInstance/TableInstance.tsx
```diff
@@ -58,6 +64,30 @@ const TableInstance: FC<ITableInstanceProps> = ({ ...props }) => {
...
+
+  const selectGroup = () => {
+    if (checked) {
+      /* if (indeterminate) {
+        const tmp = instances.filter(i => !selectiveDestroy?.includes(i.id));
+        tmp.forEach(i => selectToDestroy!(i.id));
+      } else {
+        const tmp = instances.filter(i => selectiveDestroy?.includes(i.id));
+        tmp.forEach(i => selectToDestroy!(i.id));
+      } */
```
QcFe, line 82: Remember to purge this if unused :)

frontend/src/components/activePage/TableInstance/TableInstance.tsx
```diff
@@ -85,8 +115,12 @@ const TableInstance: FC<ITableInstanceProps> = ({ ...props }) => {
                 <RowInstanceHeader
                   viewMode={viewMode}
                   handleSorting={handleSorting!}
+                  showCheckbox={showCheckbox!}
```
QcFe, line 118: Avoid skipping null checks, possibly change the prop definition

frontend/src/components/activePage/TableInstance/TableInstanceLogic.tsx
```diff
@@ -107,8 +109,17 @@ const TableInstanceLogic: FC<ITableInstanceLogicProps> = ({ ...props }) => {
           if (notify)
             notifyStatus(instance.status?.phase, instance, updateType);
 
-          const newItem = { ...prev };
-          setDataInstances(newItem);
+          //const newItem = { ...prev };
+          //setDataInstances(newItem);
```
QcFe, line 113: commented out code 😬

frontend/src/components/activePage/TableWorkspace/TableWorkspace.tsx
```diff
@@ -56,6 +82,19 @@ const TableWorkspace: FC<ITableWorkspaceProps> = ({ ...props }) => {
...
+    instances
+      .filter(i => selectiveDestroy.includes(i.id))
+      .forEach(instance => {
+        deleteInstanceMutation({
+          variables: {
+            tenantNamespace: instance.tenantNamespace,
+            instanceId: instance.name,
+          },
+        });
+      });
```
QcFe, line 95: Would rather show a spinner somewhere (if this operation is from a modal, the confirmation button could be set in a loading state) and make this all async, with a normal for so that it can await the single deletions

frontend/src/components/activePage/TableWorkspaceLogic/TableWorkspaceLogic.tsx
```diff
@@ -158,29 +177,39 @@ const TableWorkspaceLogic: FC<ITableWorkspaceLogicProps> = ({ ...props }) => {
...
+    ? instancesMapped.filter(
+        instance =>
+          multiStringIncludes(
+            filter,
+            instance.tenantId!,
+            instance.tenantDisplayName!
+          ) || selectiveDestroy.includes(instance.id)
       )
-    );
+    : [];
```
QcFe, line 192:
```suggestion
  const instancesFiltered = instancesMapped?.filter(
        instance =>
          multiStringIncludes(
            filter,
            instance.tenantId!,
            instance.tenantDisplayName!
          ) || selectiveDestroy.includes(instance.id)
    ) || [];
```
Doesn't this work?

frontend/src/components/activePage/Toolbox/Toolbox.tsx
```diff
@@ -0,0 +1,257 @@
...
+        >
+          Collapse
+        </Button>
+      </div>
+      <Divider type="horizontal" className="my-2 md:hidden" />
+      <div className="flex flex-col justify-start gap-2 xl:hidden">
+        <div className="flex items-center gap-2">
+          <Switch
+            checked={showAdvanced}
+            onClick={e => setShowAdvanced(e)}
```
QcFe, line 71:
```suggestion
            onClick={setShowAdvanced}
```

frontend/src/components/activePage/Toolbox/Toolbox.tsx
```diff
@@ -0,0 +1,257 @@
...
+            size="small"
+          />
+          <span>Show Header</span>
+        </div>
+        <div className="flex items-center gap-2">
+          <Switch
+            checked={showCheckbox}
+            onChange={e => setShowCheckbox(e)}
```
QcFe, line 79:
```suggestion
            onChange={setShowCheckbox}
```

frontend/src/components/activePage/Toolbox/Toolbox.tsx
```diff
@@ -0,0 +1,257 @@
...
+    </div>
+  );
+  return (
+    <div className="h-full flex justify-start items-center gap-2 px-5 py-3">
+      <Search
+        allowClear
+        className="hidden sm:block"
+        placeholder="Search User"
+        onChange={event => setSearchField(event.target.value)}
+        onSearch={value => setSearchField(value)}
```
QcFe, line 121:
```suggestion
        onSearch={setSearchField}
```
Check other occurrences :)

frontend/src/components/activePage/Toolbox/Toolbox.tsx
```diff
@@ -0,0 +1,257 @@
...
+      <Tooltip title="Collapse All" placement="top">
+        <Button
+          className="sm:hidden"
+          type="dark"
+          shape="circle"
+          size="middle"
+          icon={<FullscreenExitOutlined />}
+          onClick={() => setCollapseAll(true)}
+        />
+      </Tooltip> */}
```
QcFe, line 189: Is this still needed?

frontend/src/components/activePage/Toolbox/Toolbox.tsx
```diff
@@ -0,0 +1,257 @@
...
+      <Divider type="vertical" className="hidden xl:block" />
+      <div className="flex flex-col justify-start gap-2 hidden xl:block">
+        <div className="flex items-center w-36 gap-2">
+          <Switch
+            checked={showAdvanced}
+            onClick={e => setShowAdvanced(e)}
```
QcFe, line 195:
```suggestion
            onClick={setShowAdvanced}
```
And further occurrences

QcFe (commented):
My usual style bits, plus a suggestion if it's not too problematic, will test this all out sooner or later

frontend/src/components/activePage/TableInstance/TableInstance.tsx
cannarelladev: yes, i'm only waiting that someone else besides me tries and tests the deploy staging. only to be sure the changes hasn't created any problem

frontend/src/components/activePage/TableInstance/TableInstance.tsx
cannarelladev: a solution like `showCheckbox || false` could be ok?

frontend/src/components/activePage/TableWorkspace/TableWorkspace.tsx
cannarelladev: ok, i need to think the better way to do this, i have some doubts about how to wait the end of all deletions. in the other hand i have for each deletion the loading state

frontend/src/components/activePage/TableInstance/TableInstance.tsx
QcFe: Yes, actually I just noticed it's a boolean, probably !!showCheckbox is even better. The other props which aren't booleans are more problematic actually but it's out of scope now I guess

frontend/src/components/activePage/TableWorkspace/TableWorkspace.tsx
QcFe: The mutation should return a promise, in that case you should be able to await for such promise. You can't await within a forEach though, you need a regular for...

frontend/src/components/activePage/TableWorkspaceLogic/TableWorkspaceLogic.tsx
cannarelladev: it should have to

frontend/src/components/activePage/TableWorkspace/TableWorkspace.tsx
QcFe: In general the issue is that the forEach starts a possibly high number of transactions, with an api call for each of those, which could result in a bad user experience and a rather heavy load on the server side. Handling them one by one instead could be better (I think)...

frontend/src/components/activePage/TableInstance/TableInstance.tsx
cannarelladev: Ok, thx

frontend/src/components/activePage/TableWorkspace/TableWorkspace.tsx
cannarelladev: i think the same, if i remember right we already talk about this arg. so i transform it in a For sandard cycle so that i can wait on each cycle the promise (this was my doubt before)

cannarelladev:
@QcFe maybe we are ready

frontend/src/components/activePage/TableInstance/TableInstance.tsx
```diff
@@ -20,12 +20,15 @@ export interface ITableInstanceProps {
...
   showAdvanced?: boolean;
+  showCheckbox?: boolean;
   handleSorting?: (sortingType: string, sorting: number) => void;
   handleManagerSorting?: (
     sortingType: string,
     sorting: number,
     sortingTemplate: string
   ) => void;
+  selectiveDestroy?: string[];
+  selectToDestroy?: (instanceId: string) => void;
```
QcFe, line 31: Does this have to be optional?

frontend/src/components/activePage/TableInstance/TableInstance.tsx
```diff
@@ -58,6 +64,23 @@ const TableInstance: FC<ITableInstanceProps> = ({ ...props }) => {
...
+  const indeterminate = selectedIn.length !== instances.length && checked;
+
+  const selectGroup = () => {
+    if (checked) {
+      instances
+        .filter(i => indeterminate !== selectiveDestroy?.includes(i.id))
+        .forEach(i => selectToDestroy!(i.id));
+    } else {
+      instances.forEach(i => selectToDestroy!(i.id));
+    }
```
QcFe, line 81:
```suggestion
  const selectGroup = () => {
    if (!selectToDestroy) return; // if it has to be optional...
    let instIds = instances.map(({id}) => id);
    if (checked)
      instIds.filter(i => indeterminate !== selectiveDestroy?.includes(i))
    instIds.forEach(selectToDestroy);
```

frontend/src/components/activePage/TableWorkspace/TableWorkspace.tsx
```diff
@@ -56,6 +82,18 @@ const TableWorkspace: FC<ITableWorkspaceProps> = ({ ...props }) => {
...
+  const destroySelected = async () => {
+    const selection = instances.filter(i => selectiveDestroy.includes(i.id));
+    for (let i = 0; i < selection.length; i++) {
+      await deleteInstanceMutation({
+        variables: {
+          tenantNamespace: selection[i].tenantNamespace,
+          instanceId: selection[i].name,
+        },
+      });
+    }
```
QcFe, line 94:
```suggestion
    for (const { tenantNamespace, name: instanceId } of selection) {
      await deleteInstanceMutation({
        variables: { tenantNamespace, instanceId },
      });
    }
```

frontend/src/components/activePage/TableWorkspaceLogic/TableWorkspaceLogic.tsx
```diff
@@ -148,8 +160,15 @@ const TableWorkspaceLogic: FC<ITableWorkspaceLogicProps> = ({ ...props }) => {
...
             notifyStatus(instance.status?.phase, instance, updateType);
 
-          const newItem = { ...prev };
-          setDataInstances(newItem);
+          if (
+            objType === SubObjType.Addition ||
+            objType === SubObjType.Deletion ||
+            objType === SubObjType.PrettyName ||
+            objType === SubObjType.UpdatedInfo
+          ) {
```
QcFe, line 168:
```suggestion
          if (objType !== SubObjType.Drop) {
```
I saw this check in many other places... I wonder if it can't be simplified in this way everywhere...

QcFe (approved):
A couple more bits :)

frontend/src/components/activePage/TableInstance/TableInstance.tsx
cannarelladev: Yes, unfortunately because the dashboard also use the `TableInstance` without the selective destroy

PR #721, opened 2022-01-11
Add Resource Quota exceeded phase in Instance Status 

pierlagreca:
This PR adds a new environment phase in the instance status which occurs during its creation when resource quota are exceeded


operators/api/v1alpha2/instance_types.go
```diff
@@ -31,6 +31,8 @@ const (
 	EnvironmentPhaseImporting EnvironmentPhase = "Importing"
 	// EnvironmentPhaseStarting -> the environment is starting.
 	EnvironmentPhaseStarting EnvironmentPhase = "Starting"
+	// EnvironmentPhaseResourceQuotaExceeded -> the environment could not start because the resource quota are exceeded.
+	EnvironmentPhaseResourceQuotaExceeded EnvironmentPhase = "ResourceQuotaExceeded"
```
QcFe, line 35: You need to add this also in the enum above (line 21)

pierlagreca: Thanks, I will add it in the next commit

QcFe: resolve this conversation when you're done :)

frontend/src/utils.ts
```diff
@@ -34,6 +34,7 @@ export type VmStatus =
   | '' //the environment phase is unknown.
   | 'Importing' //the image of the environment is being imported.
   | 'Starting' //the environment is starting.
+  | 'ResourceQuotaExceeded' //the environment could not start because the resource quota are exceeded.
```
giorio94, line 37:
```suggestion
  | 'ResourceQuotaExceeded' //the environment could not start because the resource quota is exceeded.
```

operators/pkg/instctrl/statusinspection.go
```diff
@@ -110,3 +121,23 @@ func isVMIReady(vmi *virtv1.VirtualMachineInstance) bool {
 
 	return false
 }
+
+// isResourceQuotaExceededForVMI checks if VMI exceedes resource quota, depending on its conditions.
+func isResourceQuotaExceededForVMI(vmi *virtv1.VirtualMachineInstance) bool {
+	for _, condition := range vmi.Status.Conditions {
+		if condition.Type == "Synchronized" && condition.Status == corev1.ConditionFalse && condition.Reason == "FailedCreate" {
```
giorio94, line 128:
```suggestion
		if condition.Type == virtv1.VirtualMachineInstanceSynchronized && condition.Status == corev1.ConditionFalse && condition.Reason == "FailedCreate" {
```

operators/pkg/instctrl/statusinspection.go
```diff
@@ -110,3 +121,23 @@ func isVMIReady(vmi *virtv1.VirtualMachineInstance) bool {
...
+			return strings.Contains(condition.Message, "exceeded quota")
+		}
+	}
+	return false
+}
+
+// isResourceQuotaExceededForDeployment checks if deployment exceedes resource quota, depending on its conditions.
+func isResourceQuotaExceededForDeployment(deployment *appsv1.Deployment) bool {
+	for _, condition := range deployment.Status.Conditions {
+		if condition.Type == "ReplicaFailure" && condition.Status == corev1.ConditionTrue && condition.Reason == "FailedCreate" {
```
giorio94, line 138:
```suggestion
		if condition.Type == appsv1.DeploymentReplicaFailure && condition.Status == corev1.ConditionTrue && condition.Reason == "FailedCreate" {
```

operators/pkg/instctrl/statusinspection_test.go
```diff
@@ -31,6 +31,18 @@ import (
...
 var _ = Describe("Status Inspection", func() {
 
+	ForgeResourceQuotaExceededVMI := func() *virtv1.VirtualMachineInstance {
+		return &virtv1.VirtualMachineInstance{Status: virtv1.VirtualMachineInstanceStatus{
+			Phase: virtv1.Pending,
+			Conditions: []virtv1.VirtualMachineInstanceCondition{{
+				Type:    "Synchronized",
+				Status:  corev1.ConditionFalse,
+				Reason:  "FailedCreate",
+				Message: "exceeded quota",
```
giorio94, line 41: Would use a more realistic message taken from the real error

operators/pkg/instctrl/statusinspection_test.go
```diff
@@ -125,6 +139,17 @@ var _ = Describe("Status Inspection", func() {
 			}
 		}
 
+		ForgeResourceQuotaExceededDeployment := func(desired, ready int32) *appsv1.Deployment {
+			deployment := ForgeDeployment(desired, ready)
+			deployment.Status.Conditions = []appsv1.DeploymentCondition{{
+				Type:    "ReplicaFailure",
+				Status:  corev1.ConditionTrue,
+				Reason:  "FailedCreate",
+				Message: "exceeded quota",
```
giorio94, line 148: Same here

operators/pkg/instctrl/virtualmachines.go
```diff
@@ -83,7 +84,15 @@ func (r *InstanceReconciler) enforceVirtualMachine(ctx context.Context) error {
...
 
-	phase := r.RetrievePhaseFromVM(&vm)
+	vmiLookupKey := types.NamespacedName{Name: vm.ObjectMeta.Name}
+	var vmi virtv1.VirtualMachineInstance
+	var phase clv1alpha2.EnvironmentPhase
+	if err = r.Get(ctx, vmiLookupKey, &vmi); err != nil {
+		phase = r.RetrievePhaseFromVM(&vm, nil)
+	} else {
+		phase = r.RetrievePhaseFromVM(&vm, &vmi)
+	}
```
giorio94, line 94:
```suggestion
	// It is necessary to retrieve the VMI object associated with the VM (if any), to correctly detect the ResourceQuotaExceeded phase.
	// VM and VMI are characterized by the same resource name.
	vmi := virtv1.VirtualMachineInstance{ObjectMeta: forge.ObjectMeta(instance)}
	if err = r.Get(ctx, client.ObjectKeyFromObject(&vmi), &vmi); client.IgnoreNotFound(err) != nil {
		log.Error(err, "failed to retrieve virtualmachineinstance", "virtualmachineinstance", klog.KObj(&vm))
		return err
	}

	phase := r.RetrievePhaseFromVM(&vm, &vmi)
```

Your code had two main problems:
* It did not consider the case of possible errors, which should not be ignored
* The VMI lookup key was not correct (the namespace was missing, and the logic could not work).

BTW, the nil check in `RetrievePhaseFromVM` can be deleted, as no longer useful.

giorio94 (commented):
A couple of comments inline.

To ensure the phase is correctly updated also in case of persistent VMs, you need to do something like this in `controller.go` (you also need to create the appropriate `forge.InstanceNameFromLabels` function):

```go
// SetupWithManager registers a new controller for Instance resources.
func (r *InstanceReconciler) SetupWithManager(mgr ctrl.Manager, concurrency int) error {
	mgr.GetLogger().Info("setup manager")
	return ctrl.NewControllerManagedBy(mgr).
		For(&clv1alpha2.Instance{}).
		Owns(&appsv1.Deployment{}).
		Owns(&virtv1.VirtualMachine{}).
		// Here, we use Watches instead of Owns since we need to react also in case a VMI generated from a VM is updated,
		// to correctly update the instance phase in case of persistent VMs with resource quota exceeded.
		Watches(&source.Kind{Type: &virtv1.VirtualMachineInstance{}}, handler.EnqueueRequestsFromMapFunc(r.vmiToInstance)).
		WithOptions(controller.Options{
			MaxConcurrentReconciles: concurrency,
		}).
		Complete(r)
}

// vmiToInstance returns a reconcile request for the instance associated with the given VMI object.
func (r *InstanceReconciler) vmiToInstance(o client.Object) []reconcile.Request {
	if instance, found := forge.InstanceNameFromLabels(o.GetLabels()); found {
		return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: o.GetNamespace(), Name: instance}}}
	}

	return nil
}
```

giorio94 (approved):
Just a minor comment

operators/pkg/forge/labels.go
```diff
@@ -128,3 +128,10 @@ func persistentLabelValue(environmentList []clv1alpha2.Environment) string {
 	}
 	return strconv.FormatBool(false)
 }
+
+// InstanceNameFromLabels receives in input a set of labels and returns the instance name, if any.
+func InstanceNameFromLabels(labels map[string]string) (string, bool) {
+	// if labelInstanceKey is present instance will receive the associated value and found will be set to true.
+	instance, found := labels[labelInstanceKey]
+	return instance, found
```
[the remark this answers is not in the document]
pierlagreca: I already tried to return directly the result of that call, but the compiler reports wrong number of return values

frontend/src/components/activePage/TableInstance/RowInstanceActions/RowInstanceActionsPersistent.tsx
```diff
@@ -49,7 +49,8 @@ const RowInstanceActionsPersistent: FC<IRowInstanceActionsPersistentProps> = ({
     }
   };
 
-  return status === 'VmiReady' ? (
+  return status === 'VmiReady' ||
+    (status === 'ResourceQuotaExceeded' && running) ? (
```
giorio94, line 53: The instance is always running if the status is `ResourceQuotaExceeded`

frontend/src/components/activePage/TableInstance/RowInstanceActions/RowInstanceActionsPersistent.tsx
```diff
@@ -70,7 +71,8 @@ const RowInstanceActionsPersistent: FC<IRowInstanceActionsPersistentProps> = ({
         onClick={() => mutateInstanceStatus(false)}
       />
     </Tooltip>
-  ) : status === 'VmiOff' ? (
+  ) : status === 'VmiOff' ||
+    (status === 'ResourceQuotaExceeded' && !running) ? (
```
giorio94, line 75: Same here, the additional check should not be necessary.

frontend/src/components/activePage/TableInstance/RowInstanceActions/RowInstanceActionsPersistent.tsx
```diff
@@ -49,7 +49,8 @@ const RowInstanceActionsPersistent: FC<IRowInstanceActionsPersistentProps> = ({
     }
   };
 
-  return status === 'VmiReady' ? (
+  return status === 'VmiReady' ||
```
giorio94, line 52: Should convert to the appropriate enum, since it is now defined in the generated-types.

frontend/src/components/workspaces/Templates/TemplatesTableRow/TemplatesTableRow.tsx
```diff
@@ -92,8 +92,7 @@ const TemplatesTableRow: FC<ITemplatesTableRowProps> = ({ ...props }) => {
       .catch(() => setCreateDisabled(false));
   };
 
-  //TODO this value will be retrieved with a query
-  const InstancesLimit = 5;
+  const InstancesLimit = data?.tenant?.status?.quota?.instances!;
```
giorio94, line 95: Not an expert of typescript, but would handle the case that value is not present (e.g., falling back to a default), instead of asserting that it is always present. @QcFe WDYT?

operators/pkg/instctrl/controller.go
```diff
@@ -101,6 +104,11 @@ func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 	ctx, _ = clctx.InstanceInto(ctx, &instance)
 	tracer.Step("retrieved the instance")
 
+	// Skip the Reconcile if the instance has exceeded resource quota.
+	if instance.Status.Phase == clv1alpha2.EnvironmentPhaseResourceQuotaExceeded {
+		return ctrl.Result{}, nil
+	}
+
```
giorio94, line 111: This should be reverted, as it does not lead to the desired outcome.

operators/pkg/forge/labels.go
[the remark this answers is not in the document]
giorio94: You are right, since that can return either one or two values depending on how it is assigned.

frontend/src/components/activePage/TableInstance/RowInstanceActions/RowInstanceActionsPersistent.tsx
pierlagreca: After the call to `mutateInstanceStatus` the value of `running` boolean flag changes.
I already tried without the second check, the result is that when the user taps the stop icon in the UI instance is correctly stopped, but the icon remains the same, instead of being changed in a start icon, so the user has no feedback about this operation

giorio94: Then there is something wrong with the operator, because the phase should become Off. That is probably related to the ignore if quota exceeded part, which indeed does not work.

pierlagreca: If the correct behavior of the operator is to set the phase to off I will fix this from tha backend

frontend/src/utils.ts
```diff
@@ -34,6 +34,7 @@ export type VmStatus =
   | '' //the environment phase is unknown.
   | 'Importing' //the image of the environment is being imported.
   | 'Starting' //the environment is starting.
+  | 'ResourceQuotaExceeded' //the environment could not start because the resource quota is exceeded.
```
QcFe, line 37: I would completely remove the VmStatus type, since the new enum has been defined. By removing this type, you'll find out all the other occurrences.

QcFe (approved):

PR #731, opened 2022-01-15
Adding configuration of crownlabs user in cloud-init

ermete97:
# Description
---
This PR adds directives to cloud-init in order to create in every VM an user with username "crownlabs" and password "crownlabs".

### DIFF
--- a/operators/pkg/forge/cloudinit.go
+++ b/operators/pkg/forge/cloudinit.go
@@ -22,10 +22,21 @@ import (
 
 // userdata is a helper structure to marshal the userdata configuration.
 type userdata struct {
+	Users             []user      `yaml:"users"`
 	Network           network     `yaml:"network"`
 	Mounts            [][]string  `yaml:"mounts"`
 	WriteFiles        []writefile `yaml:"write_files"`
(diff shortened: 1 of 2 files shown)


operators/pkg/forge/cloudinit.go
```diff
@@ -49,6 +58,12 @@ type writefile struct {
 // CloudInitUserData forges the yaml manifest representing the cloud-init userdata configuration.
 func CloudInitUserData(nextcloudBaseURL, webdavUsername, webdavPassword string, publicKeys []string) ([]byte, error) {
 	config := userdata{
+		Users: []user{{
```
giorio94, line 61: Should set the default shell as bash, as it seems to use `/bin/sh` by default

giorio94, line 61: User creation has somewhat broken ssh authorized keys configuration (both in case the user does already exists and in case it does not yet). Should be checked why these are always added to the root user.

giorio94 (commented):
Some initial comments based on tests with existing VMs. Should also fill the description of the PR.

giorio94 (approved):

giorio94:
Please, squash the commits before merging

PR #732, opened 2022-01-16
Instance terminator controller

QcFe:
# Description

This PR adds the first part of logic for managing instance automatic termination given customization URLs.
Implementation consists of a controller that periodically checks instances by re-queuing them.


frisso (approved):
Approving just in case this is urgent; Marco is away a few days, not sure he'll be able to review this proptly.

QcFe:
Thanks @frisso, however maybe it's better to wait for Marco's review. We still have some time and this is quite a delicate component. My inexperience could be problematic...

operators/api/v1alpha2/instance_types.go
```diff
@@ -92,6 +92,18 @@ type InstanceSpec struct {
 	CustomizationUrls *InstanceCustomizationUrls `json:"customizationUrls,omitempty"`
 }
 
+// InstanceAutomationStatus reflects the status of the instance's automation (termination and submission).
+type InstanceAutomationStatus struct {
+	// The last time the Instance desired status was checkedd.
+	LastCheckTime metav1.Time `json:"lastCheckTime"`
+
+	// The termination time of the Instance.
+	TerminationTime metav1.Time `json:"terminationTime"`
```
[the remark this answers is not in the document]
QcFe: Actually I'd set this field only upon effective termination, not really putting the expected time from the response of the statusCheck

operators/api/v1alpha2/instance_types.go
```diff
@@ -116,6 +128,9 @@ type InstanceStatus struct {
 	// The amount of time the Instance required to become ready for the first time
 	// upon creation.
 	InitialReadyTime string `json:"initialReadyTime,omitempty"`
+
+	// Timestamps of the Instance automation phases (check, termination and submission).
+	InstanceAutomationStatus `json:",inline"`
```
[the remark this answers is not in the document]
QcFe: This is because it's inline. Is it better to make a nested object?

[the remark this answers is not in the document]
giorio94: Missed that, sorry. I would make it nested, to better group those parameters which make sense only in case of exam instances

operators/api/v1alpha2/instance_types.go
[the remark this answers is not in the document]
giorio94: I personally feel it may be useful to set it in response to the statusCheck, to have a visual idea of when the instances will e terminated (without having to dig into the logs)

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+
+	statusCheckReponse := StatusCheckResponse{}
+	statusCode, err := utils.HTTPGetJSONIntoStruct(ctx, statusCheckURL, &statusCheckReponse)
+	if statusCode != http.StatusNotFound && err != nil {
+		return false, err
+	}
+
+	status := instance.Status
+	status.Automation.LastCheckTime = metav1.Now()
+	if statusCode == http.StatusOK {
```
giorio94, line 153: Would set to now in case of NotFound error

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+	if instance.Spec.CustomizationUrls == nil {
+		return false, errors.New("customization urls field is not set for Instance")
+	}
+
+	statusCheckURL := instance.Spec.CustomizationUrls.StatusCheck
+	if statusCheckURL == "" {
+		return false, errors.New("status check url field is not set for Instance")
+	}
+
+	log := ctrl.LoggerFrom(ctx).WithName("statusCheck")
```
giorio94, line 142:
```suggestion
	log := ctrl.LoggerFrom(ctx).WithName("status-check")
```

Double-check the style of the other log messages before applying

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+func (r *InstanceTerminationReconciler) TerminateInstance(ctx context.Context, instance *clv1alpha2.Instance) error {
+	log := ctrl.LoggerFrom(ctx).WithName("termination")
+	log.Info("terminating instance")
+
+	instance = &clv1alpha2.Instance{ObjectMeta: forge.ObjectMeta(instance)}
+	op, err := ctrl.CreateOrUpdate(ctx, r.Client, instance, func() error {
+		labels := instance.GetLabels()
+		labels[forge.InstanceTerminatorSelectorLabel] = strconv.FormatBool(false)
+		labels[forge.InstanceSubmitterSelectorLabel] = strconv.FormatBool(true)
+		instance.SetLabels(labels)
```
giorio94, line 184: Would create a function in the forge package

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+		instance.Spec.Running = false
+		return nil
+	})
+
+	status := instance.Status
+	status.Automation.TerminationTime = metav1.Now()
+	if err := r.PatchStatus(ctx, instance, &status); err != nil {
+		log.Error(err, "failed updating instance status")
+		return err
+	}
```
giorio94, line 194: Could be removed, as already done as a previous step

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if err != nil {
+		log.Error(err, "failed updating Instance")
+	} else {
+		log.Info("stopped Instance", "operation", op)
+	}
```
giorio94, line 200: Should be moved above

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+	log := ctrl.LoggerFrom(ctx).WithName("statusCheck")
+	log.Info("performing instance status check")
+
+	statusCheckReponse := StatusCheckResponse{}
+	statusCode, err := utils.HTTPGetJSONIntoStruct(ctx, statusCheckURL, &statusCheckReponse)
+	if statusCode != http.StatusNotFound && err != nil {
+		return false, err
+	}
+
+	status := instance.Status
```
giorio94, line 151:
```suggestion
	status := instance.Status.DeepCopy()
```

Better to be on the safe side

operators/pkg/utils/network.go
```diff
@@ -0,0 +1,73 @@
...
+	"io"
+	"net/http"
+	"time"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	// HTTPRequestsTimeout is the timeout for HTTP requests.
+	HTTPRequestsTimeout = 5 * time.Second
```
giorio94, line 30: Should be probably lowered a bit, but it is better to also check the actual delays

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+}
+
+// TerminateInstance terminates the Instance.
+func (r *InstanceTerminationReconciler) TerminateInstance(ctx context.Context, instance *clv1alpha2.Instance) error {
+	log := ctrl.LoggerFrom(ctx).WithName("termination")
+	log.Info("terminating instance")
+
+	instance = &clv1alpha2.Instance{ObjectMeta: forge.ObjectMeta(instance)}
+	op, err := ctrl.CreateOrUpdate(ctx, r.Client, instance, func() error {
```
giorio94, line 180: Better to avoid CreateOrUpdate here, and use Update directly since:
* The instance exists for sure (and shouldn't be recreated if deleted in the meantime)
* The get operation carried out by the CreateOrUpdate function might cause a conflict, since you are possibly reading old data.

giorio94 (commented):
A few comments inline, but the overall logic seems ok to me

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	clv1alpha2 "github.com/netgroup-polito/CrownLabs/operators/api/v1alpha2"
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/forge"
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/utils"
+)
+
+const (
+	// InstanceStatusCheckRequeueTime -> time to wait before requeueing the reconcile.
+	InstanceStatusCheckRequeueTime = time.Minute * 5
```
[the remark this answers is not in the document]
QcFe: This would make sense. I was actually wondering if it's not better to make this configurable...

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,218 @@
...
+	// Check the selector label, in order to know whether to perform or not reconciliation.
+	if proceed, err := utils.CheckSelectorLabel(ctx, r.Client, instance.GetNamespace(), r.NamespaceWhitelist.MatchLabels); !proceed {
+		// If there was an error while checking, show the error and try again.
+		if err != nil {
+			log.Error(err, "failed checking selector labels")
+		}
+		return ctrl.Result{}, err
+	}
+
+	requeueResultOnError := ctrl.Result{RequeueAfter: InstanceStatusCheckRequeueTimeOnError}
```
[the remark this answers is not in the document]
QcFe: Three times now, should I still remove it?

operators/pkg/instautoctrl/terminator.go
QcFe: Isn't it better to use a Patch then? Wouldn't an Update fail in case anyone else changed the Instance in the meantime?

operators/pkg/instautoctrl/terminator.go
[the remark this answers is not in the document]
giorio94: Yep, making it configurable would be best. Would do the same also for the HTTP get timeout

operators/pkg/instautoctrl/terminator.go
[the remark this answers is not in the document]
giorio94: If the linter does not complain, I would personally repeat that. Still, as you prefer

operators/pkg/instautoctrl/terminator.go
giorio94: Would personally avoid using patch whenever possible, as it is associated with higher complexity and possible pitfalls (would be better to remove that from the status update too). If everything is written correctly, you should not incur in self-update conflicts, and the rate of other updates is definitely low, so that should not lead to problems. Still, in case of conflicts it is better to retry immediately, instead of waiting for the 5 seconds requeue time.

giorio94: Would personally avoid using patch whenever possible, as it is associated with higher complexity and possible pitfalls (would be better to remove that from the status update too). If everything is written correctly, you should not incur in self-update conflicts, and the rate of other updates is definitely low, so that should not lead to problems. Still, in case of conflicts it is better to retry immediately, instead of waiting for the 5 seconds requeue time.

QcFe: Don't know if I did it in the right way

giorio94: That makes sense to me. (Could be optimized checking if an update is actually necessary, but that might lead to possible inconsistencies, in case the status is updated at the same time. BTW, that might also be a reason for the bug in the instance controller if still present)

giorio94 (commented):
A couple of small comments

operators/pkg/instautoctrl/terminator.go
QcFe: The current setup seems to be functional. Is there anything else to be done?

operators/pkg/utils/common.go
```diff
@@ -59,7 +56,7 @@ func CheckSelectorLabel(ctx context.Context, k8sClient client.Client, namespaceN
 			return false, nil
 		}
 	} else {
-		return false, fmt.Errorf("error when retrieving the InstanceSnapshot namespace -> %w", err)
+		return false, fmt.Errorf("error when retrieving namespace <%s>: %w", namespaceName, err)
```
giorio94, line 59:
```suggestion
		return false, fmt.Errorf("error when retrieving namespace %q: %w", namespaceName, err)
```

That takes care of automatic quoting

operators/pkg/utils/network.go
```diff
@@ -0,0 +1,68 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// HTTPGet performs a HTTP GET request to the given URL.
+func HTTPGet(ctx context.Context, url string, timeout time.Duration) (statusCode int, contents []byte, err error) {
+	now := time.Now()
+	log := ctrl.LoggerFrom(ctx, "url", url).V(LogDebugLevel).WithName("http-get")
+	defer func() {
+		log.Info("terminated", "duration", time.Since(now))
+	}()
```
giorio94, line 34: It is probably more appropriate to create a sub span of  the tracer here (you can create a different LogIfLong function if necessary)

operators/pkg/instautoctrl/terminator.go
```diff
@@ -0,0 +1,177 @@
...
+
+	clv1alpha2 "github.com/netgroup-polito/CrownLabs/operators/api/v1alpha2"
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/forge"
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/utils"
+)
+
+const (
+	// InstanceStatusCheckRequeueTimeOnError -> time to wait before requeueing the reconcile on error.
+	InstanceStatusCheckRequeueTimeOnError = time.Second * 5
+)
```
giorio94, line 43: Seems unused to me

giorio94 (approved):

operators/pkg/instautoctrl/terminator.go
giorio94: I would say not

PR #737, opened 2022-01-21
Exams/Exercises chore refactors

QcFe:
# Description

This PR includes several small changes in component locations and some cleanup.

- Change exercises.* and exams.* subdomains for instance ingresses to exercise.* and exam.*
- Rename folders for instance automation utilities
- Remove the old instanceset landing (used for September exams session)


provisioning/containers/utils/content-uploader/entrypoint.sh
```diff
@@ -20,4 +20,4 @@ zip "/tmp/$FILENAME.zip" -r "$SOURCE_PATH"
 
 
 echo "Uploading archive..."
-curl -v -X POST -F "binfile=@\"/tmp/$FILENAME.zip\"" -F "filename=$FILENAME.zip" "$DESTINATION_URL"
+curl -v --fail-with-body --request POST --form "binfile=@\"/tmp/$FILENAME.zip\"" --form "filename=$FILENAME.zip" "$DESTINATION_URL"
```
giorio94, line 23: `fail-with-body` seems a non-existent option to me

giorio94 (approved):

provisioning/containers/utils/content-uploader/entrypoint.sh
QcFe: Obviously I saw it [here](https://curl.se/docs/manpage.html#--fail-with-body) but I haven't checked if it actually worked

PR #755, opened 2022-02-16
Normalize Instance status phase

QcFe:
# Description

This PR removes the old VMI text from Instance status.


operators/api/v1alpha2/instance_types.go
```diff
@@ -38,13 +38,13 @@ const (
 	// EnvironmentPhaseReady -> the environment is ready to be accessed.
 	// TODO: update the value to Ready once no more needed for backward compatibility
```
giorio94, line 39: Should delete the TODO. Same below

giorio94 (approved):

PR #762, opened 2022-03-07
Add controller for counting cpu used by Tenant

pierlagreca:
This PR adds a new controller that computes the amount of cpu minutes usage for each Tenant, and a minutes counter in the Tenant object.
This controller is bound to the `Tenant Operator` and it is scheduled to be executed every 5 minutes, when in execution the controller iterates on each instance and increases the token consumption counter of the tenant by an amount of seconds for each cpu core used, depending on last esecution of the controller itself or the creation timestamp of the instances.


pierlagreca:
deploy-staging

PR #768, opened 2022-03-17
Updated Templates Samples

cheina97:
This PR updates the examples in samples with the possibility to create a template with the marketplace disabled. I removed the example about enabling the internal proxy in code-server because I've noticed that with the latest version it is enabled by default and there is no way to disable it.

### DIFF
--- a/operators/samples/templates.yaml
+++ b/operators/samples/templates.yaml
@@ -63,18 +63,17 @@ metadata:
   name: vscode-python-proxyenabled
   namespace: workspace-standalone
 spec:
-  prettyName: Visual Studio Code - Python (proxy enabled)
+  prettyName: Visual Studio Code - Python (marketplace disabled)
   description: A template about vscode
   environmentList:
     - name: vscode-environment
       environmentType: Standalone
       mode: Standard
-      image: harbor.crownlabs.polito.it/crownlabs-standalone/vscode-python:1.0.0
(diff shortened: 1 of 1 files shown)


operators/samples/templates.yaml
```diff
@@ -73,8 +73,7 @@ spec:
       containerStartupOptions:
         contentPath: /config/workspace
         startupArgs:
-          - "false"
-          - "proxy"
+          - "true"
```
giorio94, line 76: Would personally make this argument more self explanatory, like `--disable-marketplace`, otherwise it is unclear its meaning.

giorio94 (approved):

operators/samples/templates.yaml
cheina97: `for ARGUMENT in "$@"; do
    key=$(echo "$ARGUMENT"|cut -d = -f 1)
    value=$(echo "$ARGUMENT"|cut -d = -f 2)

    if [ "$key" == "--marketplace-disabled" ] && [ "$value" == "true" ]; then
        export EXTENSIONS_GALLERY='{"serviceUrl": ""}'
    fi
done`

Could be a good idea to add something like that inside the container entry point or do you know any better solutions?

giorio94: Would just use `--disable-marketplace` (without true/false) to avoid the complexity of spitting the argument, but as you prefer.

PR #774, opened 2022-04-02
Remove the OPA policies folder

giorio94:
# Description

This PR drops the OPA policies folder, since those policies are currently unused and they have been replaced by a custom webhook. In addition, it bumps the golanci-lint action, as affected by a problem.


QcFe (commented):
Was wondering if it's not the case to make an "archive" folder or something similar, in order to not just trash everything...

giorio94:
> Was wondering if it's not the case to make an "archive" folder or something similar, in order to not just trash everything...

I see your point, although I am personally a bit against it to keep the repo as lean as possible:
* Git is already the archive, and in future everything could be restored from there;
* This originated from a very interesting project and we learned a lot. Still, I sincerely doubt we will ever reuse a solution like this in CrownLabs as it is way too complex and led to different problems (it is much simpler and flexible to write custom webhooks).

QcFe:
> I see your point, although I am personally a bit against it to keep the repo as lean as possible:
> 
> * Git is already the archive, and in future everything could be restored from there;
> * This originated from a very interesting project and we learned a lot. Still, I sincerely doubt we will ever reuse a solution like this in CrownLabs as it is way too complex and led to different problems (it is much simpler and flexible to write custom webhooks).

Sorry for the late reply.
This definitely makes sense, however these removed pieces would end up lost between commits... What if we made a kind of "features history log" to be put in documentation or somewhere else, with a list of removed features pointing at the latest tag that included them?

QcFe (approved):

PR #775, opened 2022-04-06
Removing storybook dependencies, files and workflows

cannarelladev:
# Description

In this PR storybook is definitely remoded:
- Build and workflows
- Dependencies
- Story files


cannarelladev:
@giorio94 could you check if everything is right on latest commit of this PR? the other commits belong to and old PR that is going to be merged

PR #784, opened 2022-04-28
Remove cpu maximum constraint from Template crd

QcFe:
# Description

This removes the CPU maximum constraint in the Environment spec of the Template CRD since now limits are better handled by the Tenant.

### DIFF
--- a/operators/api/v1alpha2/template_types.go
+++ b/operators/api/v1alpha2/template_types.go
@@ -123,10 +123,9 @@ type Environment struct {
 // (i.e. CPU, RAM, ...) assigned to a certain environment.
 type EnvironmentResources struct {
 	// +kubebuilder:validation:Minimum:=1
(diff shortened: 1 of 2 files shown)


operators/api/v1alpha2/template_types.go
```diff
@@ -122,9 +122,6 @@ type Environment struct {
 // EnvironmentResources is the specification of the amount of resources
 // (i.e. CPU, RAM, ...) assigned to a certain environment.
 type EnvironmentResources struct {
-	// +kubebuilder:validation:Minimum:=1
-	// +kubebuilder:validation:Maximum:=8
```
giorio94, line 126: Could make sense to keep at least the check that the number is non-negative (or leave the minimum set to one), to prevent strange failures.

giorio94 (approved):

PR #786, opened 2022-05-08
Used ResourceQuotas in Tenant Status

cheina97:
This PR adds the possibility to watch the ResourceQuota usage from the tenant status.

ToDo:

- [x] Update Tenant CRD
- [x] Run and Put correct information inside the status
- [x] Correct Old tests
- [x] Cleanup
- [x] Write forge tests

**instance operators' main.go has been temporarily modified to run instance and tenant operators at the same time**


operators/api/v1alpha2/tenant_types.go
```diff
@@ -82,11 +82,11 @@ type TenantSpec struct {
 	CreateSandbox bool `json:"createSandbox,omitempty"`
 
 	// The amount of resources associated with this Tenant, if defined it overrides the one computed from the workspaces the tenant is enrolled in.
-	Quota *TenantResourceQuota `json:"quota,omitempty"`
+	QuotaLimits *TenantResourceQuotaData `json:"quota,omitempty"`
```
giorio94, line 85: Would keep the Quota field name, for consistency with the json tag.

operators/api/v1alpha2/tenant_types.go
```diff
@@ -82,11 +82,11 @@ type TenantSpec struct {
...
 }
 
-// TenantResourceQuota defines resource quota for each Tenant.
-type TenantResourceQuota struct {
+// TenantResourceQuotaData defines resource quotas details.
```
giorio94, line 88:
```suggestion
// TenantResourceQuotaData defines resource quota details.
```

operators/cmd/instance-operator/main.go
```diff
@@ -61,7 +61,7 @@ func main() {
 	svcUrls := instctrl.ServiceUrls{}
 	instSnapOpts := instancesnapshot_controller.ContainersSnapshotOpts{}
 
-	metricsAddr := flag.String("metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	metricsAddr := flag.String("metrics-addr", ":8088", "The address the metric endpoint binds to.")
```
giorio94, line 64: Should be reverted, same below

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -320,22 +327,29 @@ func (r *TenantReconciler) createOrUpdateClusterResources(ctx context.Context, t
 	}
 
 	var retErr error
+	// retrieve instances
+	instances := crownlabsv1alpha2.InstanceList{}
+	err = r.Client.List(ctx, &instances, &client.ListOptions{Namespace: nsName})
+	if err != nil {
+		klog.Errorf("Unable to retrieve instances")
+		retErr = err
+	}
```
giorio94, line 336: Why are you retrieving the instances? All information should be present in the resource quota status.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -320,22 +327,29 @@ func (r *TenantReconciler) createOrUpdateClusterResources(ctx context.Context, t
...
 	// handle resource quota
 	rq := v1.ResourceQuota{
 		ObjectMeta: metav1.ObjectMeta{Name: "crownlabs-resource-quota", Namespace: nsName},
 	}
 	rqOpRes, err := ctrl.CreateOrUpdate(ctx, r.Client, &rq, func() error {
 		rq.Labels = r.updateTnResourceCommonLabels(rq.Labels)
-		rq.Spec.Hard = forge.TenantResourceQuotaSpec(&tn.Status.Quota)
-
+		rq.Spec.Hard = forge.TenantResourceQuotaSpec(tn.Status.Quota.Limits)
+		tn.Status.Quota.Used = forge.TenantResourceQuotaStatusUsed(&rq, &instances)
```
giorio94, line 344: This is the resource quota mutation function, and should be used to mutate the resource quota, and not extract additional values. That can be done as a subsequent step accessing the `rq` value.

operators/api/v1alpha2/tenant_types.go
```diff
@@ -98,6 +98,14 @@ type TenantResourceQuota struct {
 	Instances uint32 `json:"instances"`
 }
 
+// TenantResourceQuota defines maximum and current resource quota for each Tenant.
+type TenantResourceQuota struct {
+	// The amount of used resources associated with this Tenant.
+	Used *TenantResourceQuotaData `json:"used,omitempty"`
+	// The amount of limits resources available to this Tenant.
+	Limits *TenantResourceQuotaData `json:"limits,omitempty"`
```
giorio94, line 106: These are not required to be pointers (pointers are typically used in APIs to distinguish whether a field has not been set or has been assigned the zero value, but in this case it is not necessary and adds a bit of complexity).

operators/pkg/tenant-controller/tenant_controller_test.go
```diff
@@ -311,6 +311,8 @@ func checkTnClusterResourceCreation(ctx context.Context, tnName, nsName string,
 	Expect(createdRq.Spec.Hard["requests.memory"]).Should(Equal(reqMem))
 	Expect(createdRq.Spec.Hard["count/instances.crownlabs.polito.it"]).Should(Equal(instanceCount))
 
+	By("By checking that the tenant status used resource quotas have been correctly updated")
```
giorio94, line 314: Leveraging the resource quota only makes it also easier to add a small test here. Just manually update the resource quota status and assert that the same values are eventually reflected in the tenant status.

giorio94 (commented):
A few initial high-level comments

PR #799, opened 2022-07-12
EnforceWorkdir & Submission Jobs TTL

QcFe:
# Description

This PR includes the following changes:
- add TTLSecondsAfterFinished (fixed to 5 minutes) to Submission Jobs: in case of success or failure, after such time the Job is deleted. In case of successful submission this provides simple cleanup. In case of error, this should trigger the recreation of the Job from the submission controller.
- add the EnforceWorkdir directive to the ContainerStartupOpts field in the template. In case this directive is set to true, the application container workdir will be enforced to be the one specified by the contentPath (if set) or to the default mydrive path (otherwise). When enforcing the workdir, volume mount is ensured onto the workdir. 
- changes websockify probe from TCP to HTTP

# How Has This Been Tested?
- [x] Unit tests
- [x] Staging

### DIFF
--- a/operators/api/v1alpha2/template_types.go
+++ b/operators/api/v1alpha2/template_types.go
@@ -160,6 +160,10 @@ type ContainerStartupOpts struct {
 	ContentPath string `json:"contentPath,omitempty"`
 	// Arguments to be passed to the application container on startup
 	StartupArgs []string `json:"startupArgs,omitempty"`
+	// Whether forcing the container working directory to be the same as the contentPath (or default mydrive path if not specified)
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	EnforceWorkdir bool `json:"enforceWorkdir"`
 }
 
 // +kubebuilder:object:root=true
--- a/operators/pkg/forge/containers.go
+++ b/operators/pkg/forge/containers.go
@@ -53,6 +53,8 @@ const (
 	CrownLabsUserID = int64(1010)
 	// SubmissionJobMaxRetries -> max number of retries for submission jobs.
 	SubmissionJobMaxRetries = 10
+	// SubmissionJobTTLSeconds -> seconds for submission jobs before deletion (either failure or success).
+	SubmissionJobTTLSeconds = 300
 	// AppCPULimitsEnvName -> name of the env variable containing AppContainer CPU limits.
 	AppCPULimitsEnvName = "APP_CPU_LIMITS"
 	// AppMEMLimitsEnvName -> name of the env variable containing AppContainer memory limits.
@@ -159,7 +161,8 @@ func PodSpec(instance *clv1alpha2.Instance, environment *clv1alpha2.Environment,
 // SubmissionJobSpec returns the job spec for the submission job.
 func SubmissionJobSpec(instance *clv1alpha2.Instance, environment *clv1alpha2.Environment, opts *ContainerEnvOpts) batchv1.JobSpec {
 	return batchv1.JobSpec{
-		BackoffLimit: pointer.Int32(SubmissionJobMaxRetries),
+		BackoffLimit:            pointer.Int32(SubmissionJobMaxRetries),
+		TTLSecondsAfterFinished: pointer.Int32(SubmissionJobTTLSeconds),
 		Template: corev1.PodTemplateSpec{
 			Spec: corev1.PodSpec{
 				Containers: []corev1.Container{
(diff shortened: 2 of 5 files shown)


operators/api/v1alpha2/template_types.go
```diff
@@ -160,6 +160,9 @@ type ContainerStartupOpts struct {
 	ContentPath string `json:"contentPath,omitempty"`
 	// Arguments to be passed to the application container on startup
 	StartupArgs []string `json:"startupArgs,omitempty"`
+	// Whether forcing the container working directory to be the same as the contentPath (or default mydrive path if not specified)
+	// +kubebuilder:default=true
+	EnforceWorkdir bool `json:"enforceWorkdir,omitempty"`
```
giorio94, line 165: Do not use a boolean value defaulted to true coupled with the omitempty option, as this has shown to create issues with controllers (a false value is not transmitted, thus considered as true by the backend). In other words, I would suggest removing the omitempty option and adding the `// +kubebuilder:validation:Optional` marker.

giorio94 (approved):
/lgtm

Just perform a couple of tests in the staging environment before merging

operators/api/v1alpha2/template_types.go
QcFe: That's exactly the issue I had in mind when I wrote that, and what I wanted to avoid but, as my usual, I messed up. Thanks

QcFe: My original intent was to leave the omitempty and default it to false instead. What do you think?

giorio94: Would personally use the Optional marker instead of omitempty, regardless of the default value (feel free to select the one you prefer).

QcFe:
From staging tests, enforcing the workdir is not always the best approach: it seems that node.js for example needs the workdir to be the package root and forcing another one breaks the container. 
What I implemented to make sure the contentPath folder is mounted is to check whether the contentPath property is explicitly set. In that case the container volume will be created and mounted. 
In general however I'd keep the new enforceWorkdir property as it might happen to be useful in other possible contexts...

operators/pkg/forge/containers.go
```diff
@@ -470,7 +476,13 @@ func ContainerVolume(volumeName, claimName string, environment *clv1alpha2.Envir
 // NeedsContainerVolume returns true in the cases in which a volume mount could be needed.
 func NeedsContainerVolume(instance *clv1alpha2.Instance, environment *clv1alpha2.Environment) bool {
 	needsInit, _ := NeedsInitContainer(instance, environment)
-	return environment.Mode == clv1alpha2.ModeStandard || environment.Persistent || needsInit
+	enforceWorkDir := false
+	hasExplicitContentPath := false
+	if environment.ContainerStartupOptions != nil {
+		enforceWorkDir = environment.ContainerStartupOptions.EnforceWorkdir
+		hasExplicitContentPath = environment.ContainerStartupOptions.ContentPath != ""
```
giorio94, line 483:
```suggestion
		enforceWorkDir := environment.ContainerStartupOptions != nil && environment.ContainerStartupOptions.EnforceWorkdir
		hasExplicitContentPath := environment.ContainerStartupOptions != nil && environment.ContainerStartupOptions.ContentPath != ""
```

A bit more concise

operators/pkg/forge/containers.go
```diff
@@ -470,7 +476,13 @@ func ContainerVolume(volumeName, claimName string, environment *clv1alpha2.Envir
 // NeedsContainerVolume returns true in the cases in which a volume mount could be needed.
 func NeedsContainerVolume(instance *clv1alpha2.Instance, environment *clv1alpha2.Environment) bool {
```
giorio94, line 477: I was wondering whether it wouldn't be better to always create the container volume, for the sake of simplicity (thus dropping this function). After all, it seems there would be practically no real case in which it is not created.

giorio94:
@QcFe  Any updates on this?

QcFe:
Clearly I completely forgot about this. Pushing some update, then I'll have to test stuff out

operators/pkg/forge/containers.go
QcFe: Followed your suggestion and removed this method, always adding the volume

QcFe:
@giorio94 did a couple of tests also for submission jobs and things seem to work...

giorio94 (approved):

PR #818, opened 2022-11-17
Improve CSV upload and error handling from qlkube

QcFe:
# Description

This PR solves several issues with the uploading procedure for CSV imports and tries to make error reporting more user friendly.

Added a few reviewers just to let them know, feel free to have a look or anything.


qlkube/src/index.js
```diff
@@ -50,6 +50,28 @@ async function main() {
     schema,
     playground: true,
     plugins: [graphqlQueryRegistry],
+    formatError: error => {
+      const msgs = [...error.message.match(/(\d+)\s*\-(.*)$/)];
```
ralls0, line 54: Trying a wrong query in order to check the result obtained I get:
`"message": "error.message.match is not a function or its return value is not iterable"` instead of `"message": "Context creation failed: Token Error! Token absent!"`.
The query used is:
```
{
  itPolitoCrownlabsV1alpha1Workspace(name: "prova"){
    kind
  }
}
```

QcFe: Thanks! I guess wrapping everything in the try/catch can be more robust

PR #826, opened 2022-12-20
Nfs storage

matteorosani:
# Description

The tenant controller creates a pvc for the tenant and uses it as private storage for the user.

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B


operators/pkg/forge/cloudinit.go
```diff
@@ -75,18 +74,13 @@ func CloudInitUserData(nextcloudBaseURL, webdavUsername, webdavPassword string,
 			ID0:     interf{DHCP4: true},
 		},
 		Mounts: [][]string{{
-			fmt.Sprintf("%s/remote.php/dav/files/%s", nextcloudBaseURL, webdavUsername),
-			"/media/MyDrive",
-			"davfs",
-			"_netdev,auto,user,rw,uid=1000,gid=1000",
+			fmt.Sprintf("%s:%s", serviceName, servicePath),
+			"/media/nfs-share",
```
QcFe, line 78: Would revert the mountpoint: iirc, basically all images have a shortcut pointing to there...

operators/pkg/instctrl/cloudinit.go
```diff
@@ -45,47 +46,82 @@ const (
...
 
 	// Retrieve the public keys
-	publicKeys, err := r.GetPublicKeys(ctx)
-	if err != nil {
-		log.Error(err, "unable to get public keys")
-		return err
+	publicKeys, retErr := r.GetPublicKeys(ctx)
+	if retErr != nil {
+		log.Error(retErr, "unable to get public keys")
+		return retErr
```
QcFe, line 57: would revert this

operators/pkg/instctrl/cloudinit.go
```diff
@@ -45,47 +46,82 @@ const (
...
+	if retErr != nil {
+		log.Error(retErr, "unable to marshal secret content")
+		return retErr
 	}
 
 	// Enforce the cloud-init secret presence
 	instance := clctx.InstanceFrom(ctx)
 	secret := corev1.Secret{ObjectMeta: forge.ObjectMeta(instance)}
-	res, err := ctrl.CreateOrUpdate(ctx, r.Client, &secret, func() error {
-		secret.SetLabels(forge.InstanceObjectLabels(secret.GetLabels(), instance))
```
QcFe, line 74: Labels settings should be preserved

operators/pkg/instctrl/cloudinit.go
```diff
@@ -45,47 +46,82 @@ const (
...
 		secret.Type = corev1.SecretTypeOpaque
 		return ctrl.SetControllerReference(instance, &secret, r.Scheme)
 	})
 
-	if err != nil {
-		log.Error(err, "failed to enforce cloud-init secret", "secret", klog.KObj(&secret))
-		return err
+	if retErr != nil {
+		log.Error(retErr, "failed to enforce cloud-init secret", "secret", klog.KObj(&secret))
+		return retErr
```
QcFe, line 83: Would revert

operators/pkg/instctrl/containers_test.go
```diff
@@ -102,11 +103,21 @@ var _ = Describe("Generation of the container based instances", func() {
...
+					Name:    instanceNamespace,
+				},
+			},
+		}
 		instance = clv1alpha2.Instance{
 			ObjectMeta: metav1.ObjectMeta{Name: instanceName, Namespace: instanceNamespace},
 			Spec: clv1alpha2.InstanceSpec{
 				Running:  true,
 				Template: clv1alpha2.GenericRef{Name: templateName, Namespace: templateNamespace},
+				Tenant:   clv1alpha2.GenericRef{Name: tenantName},
```
QcFe, line 120: A test should be added for when a tenant is not set: the instance should still be able to start if the nfs share mount is not required by the template spec. (Exam instances do not use tenants, the tenant operator is not even deployed)

QcFe (commented):
Just gave a super brief look and left a couple of little things as I saw them, actual review is still to be done

operators/pkg/forge/cloudinit.go
matteorosani: Reverted mount point

operators/pkg/instctrl/cloudinit.go
matteorosani: I don't think I fully understand what you are saying. I've changed the name of the variable back to ```err``` and placed this piece of code after the get of the nfs secret like it was after the ```r.GetWebDavCredentials(ctx)```.

operators/pkg/instctrl/cloudinit.go
matteorosani: Fixed, it was an error in transcription

operators/pkg/instctrl/containers_test.go
matteorosani: In that case the ```environment.Mode`` is set to ```Standard``` right?
In that case I'm sure it doesn't mount the volume, not even searching for it.
But to be more general: how can I check that the instance has no tenant? The ```instance.Spec.Tenant``` can't be ```nil``` so the ```instance.Spec.Tenant.Name``` is empty string?

operators/pkg/instctrl/cloudinit.go
QcFe: Yes, in general we're talking about diffs of code, relatively to how it was before your commits. You can see diffs in the "files changed" tab of the PR. You can resolve conversations directly when you handled them, in case no further information/discussion is needed.

operators/pkg/instctrl/containers_test.go
QcFe: Correct. `Spec.Tenant` cannot be nil since it's a nested structure. In general we can consider the tenant absent when it's an empty string, it's a rather common approach.

operators/pkg/forge/cloudinit.go
```diff
@@ -58,36 +57,52 @@ type writefile struct {
...
+				Sudo:              "ALL=(ALL) NOPASSWD:ALL",
+				SSHAuthorizedKeys: publicKeys,
+				Shell:             "/bin/bash",
+			}},
+			Network: network{
+				Version: 2,
+				ID0:     interf{DHCP4: true},
+			},
+			SSHAuthorizedKeys: publicKeys,
+		}
```
QcFe, line 105: You're duplicating the code heavily. You can create an initial, common userdata, then, 
```go
if serviceName != "" && servicePath != "" {
   config.Mounts = ...
}

File: operators/api/v1alpha2/template_types.go [hunk 7]
```diff
@@ -122,6 +122,9 @@ type Environment struct {
 
 	// Name of the storage class to be used for the persistent volume (when needed)
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// Whether the instance has to have the user path
+	MountPathEnabled bool `json:"mountPathEnabled,omitempty"`
```
QcFe, line 127: Would not set it as `omitempty`, also would add the kubebuilder flag for defaulting it to `true`

operators/pkg/instctrl/cloudinit.go
```diff
@@ -44,14 +45,21 @@ const (
...
+	if env.MountPathEnabled {
+		_, server, path, err = r.GetNfsSpecs(ctx)
+		if err != nil {
+			log.Error(err, "unable to retrieve nfs volume dns name and path")
+			return err
+		}
+	} else {
+		server = ""
+		path = ""
 	}
```
QcFe, line 62: Not needed, new strings are already initialized as empty

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,6 +94,41 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
 	return nil
 }
 
+// Get the specs of the volume
+func (r *InstanceReconciler) GetNfsSpecs(ctx context.Context) (serverIp, serverName, path string, err error) {
+	tenant := clctx.TenantFrom(ctx)
+	secretName := types.NamespacedName{Namespace: tenant.Namespace, Name: tntctrl.PvcSecretName}
```
QcFe, line 100: I'm not sure this is really needed: given the basic CrownLabs structure, instances should be already created within the tenant namespace, this retrieving the NS from the Instance itself would be enough. @giorio94 is this correct?

operators/pkg/instctrl/containers.go
```diff
@@ -28,6 +31,60 @@ import (
...
+func (r *InstanceReconciler) GetNFSVolume(ctx context.Context, volume *v1.Volume, volumeMount *v1.VolumeMount) error {
+	// Retrieve the correct Tenant from the given context
+	tenant := clctx.TenantFrom(ctx)
+	if tenant == nil {
+		retErr := errors.New("unable to retrieve tenant from the context")
+		klog.Errorf("%s", retErr)
+		return retErr
+	}
+
+	if tenant.Status.PersonalNamespace.Created {
```
QcFe, line 52: Following the previous comment: I'd retrieve the secret directly from the same namespace of the instance, since it should already be in the same namespace of the secret.

operators/pkg/instctrl/containers_test.go
```diff
@@ -144,6 +155,7 @@ var _ = Describe("Generation of the container based instances", func() {
 			ContainerEnvOpts: containerOpts,
 		}
 
+		ctx, _ = clctx.TenantInto(ctx, &tenant)
```
QcFe, line 158: If the previous two comments are meaningful, this is not necessary.

QcFe (commented):
Some other comments

operators/pkg/instctrl/containers_test.go
matteorosani: Ok, removed

operators/pkg/instctrl/containers.go
matteorosani: Ok, resolved

operators/pkg/instctrl/cloudinit.go
matteorosani: Now using the instance namespace as it was not correct, the tenant namespace is ""

operators/pkg/instctrl/cloudinit.go
matteorosani: Didn't know, removed

operators/api/v1alpha2/template_types.go
matteorosani: Added flag for default value and generated the manifests

operators/pkg/forge/cloudinit.go
matteorosani: Removed duplicated code

operators/pkg/instctrl/cloudinit.go
QcFe: Yes, also because the Tenant is a cluster wide resource, so the Namespace field (that likely comes from Metadata.Namespace) is empty, not being namespaced. The field you were looking for is more likely inside the tenant Status, but again, I'd not look for that and stick to retrieving the secret from the same NS of the instance.

operators/pkg/instctrl/containers_test.go
matteorosani: Ok added check in the controller for empty string Tenant reference in Instance

operators/pkg/instctrl/cloudinit_test.go
```diff
@@ -102,35 +103,36 @@ var _ = Describe("Generation of the cloud-init configuration", func() {
...
 		}
 	}
 
 	JustBeforeEach(func() {
 		client := FakeClientWrapped{Client: clientBuilder.Build()}
 		reconciler = instctrl.InstanceReconciler{
-			Client: client, Scheme: scheme.Scheme, WebdavSecretName: webdavCredentials,
-			ServiceUrls: instctrl.ServiceUrls{NextcloudBaseURL: nextcloudBaseURL},
+			Client: client, Scheme: scheme.Scheme,
+			ServiceUrls: instctrl.ServiceUrls{},
```
QcFe, line 135: This could be omitted, when structs are nested without being pointers, they're implicitly set as empty already.

operators/pkg/instctrl/cloudinit.go
```diff
@@ -44,14 +44,18 @@ const (
...
-		log.Error(err, "unable to get webdav credentials")
-		return err
+	env := clctx.EnvironmentFrom(ctx)
+	var server string
+	var path string
+	var err error
+
+	if env.MountPathEnabled {
+		_, server, path, err = r.GetNfsSpecs(ctx)
+		if err != nil {
```
QcFe, line 54: IIRC, you should be able to do something alike, also removing `var err error`:
```suggestion
		if _, server, path, err := r.GetNfsSpecs(ctx); err != nil {
```

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,6 +90,41 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
...
+	instance := clctx.InstanceFrom(ctx)
+	secretName := types.NamespacedName{Namespace: instance.Namespace, Name: UserPvcSecret}
+
+	secret := corev1.Secret{}
+	if err = r.Get(ctx, secretName, &secret); err != nil {
+		ctrl.LoggerFrom(ctx).Error(err, "failed to retrieve secret", "secret", secretName)
+		return
+	}
+
+	var ok bool
```
QcFe, line 104: As above, you should be able to define local variables on the go using the `:=` assignment, avoiding to define them beforehand (referenced mostly to `var ok bool` in this case).

operators/pkg/instctrl/containers.go
```diff
@@ -28,6 +31,53 @@ import (
...
+func (r *InstanceReconciler) GetNFSVolume(ctx context.Context, volume *v1.Volume, volumeMount *v1.VolumeMount) error {
+	// Retrieve the correct Tenant from the given context
+	instance := clctx.InstanceFrom(ctx)
+	if instance == nil {
+		err := errors.New("unable to retrieve instance from the context")
+		klog.Errorf("%s", err)
+		return err
+	}
+	// Get the secret and the NFS path
+	secret := v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: UserPvcSecret, Namespace: instance.Namespace}}
```
QcFe, line 52: The following should work as well, since you're providing information through NamespacedName. Defining the ObjectMeta is necessary only for the CreateOrUpdate.
```suggestion
	secret := v1.Secret{}
```

operators/pkg/instctrl/containers.go
```diff
@@ -89,6 +139,22 @@ func (r *InstanceReconciler) enforceContainer(ctx context.Context) error {
 		// either rejected or cause the restart of the Pod, with consequent possible data loss.
 		if depl.CreationTimestamp.IsZero() {
 			depl.Spec = forge.DeploymentSpec(instance, environment, &r.ContainerEnvOpts)
+
+			if environment.Mode == v1alpha2.ModeStandard && environment.MountPathEnabled && instance.Spec.Tenant.Name != "" {
```
QcFe, line 143: I'm still not too sure this is the best approach. I'd directly try retrieval of the secret in the current instance NS, regardless of the presence of the tenant name. Also, I'd set as the only check condition, the presence of `environment.MountPathEnabled`. Special Modes could still be able to mount the shared storage...

operators/pkg/instctrl/controller_test.go
```diff
@@ -86,11 +86,24 @@ var _ = Describe("The instance-controller Reconcile method", func() {
...
 		tenant := clv1alpha2.Tenant{
 			ObjectMeta: metav1.ObjectMeta{Name: testName},
 			Spec: clv1alpha2.TenantSpec{
 				Email: "test@email.me",
 			},
+			Status: clv1alpha2.TenantStatus{
+				PersonalNamespace: clv1alpha2.NameCreated{
+					Name: testName,
+				},
+			},
```
QcFe, line 106: This should not be required anymore (since the instance should not check this), not sure though.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -46,6 +47,21 @@ import (
...
+	// PVC secret name
+	PvcSecretName = "user-pvc-secret"
+	// Server key in PVC secret
+	SecretServerNameKey = "server-name"
+	// Server key in PVC secret
+	SecretServerIPKey = "server-ip"
+	// Path key in PVC secret
+	SecretPathKey = "path"
+	// PVC size
+	UserPvcSize = 1 * 1024 * 1024 * 1024 // 1 Gi
```
QcFe, line 61: There should be a way to define this using the resource type, if not, I'd still call it like UserPvcSizeBytes. Also, it should be better to make this customizable in some way, would keep this comment open for further discussion.

operators/pkg/instctrl/cloudinit.go
Diegomangasco: Ok, removed `var err error`

operators/pkg/instctrl/cloudinit_test.go
Diegomangasco: Ok, removed useless items

operators/pkg/instctrl/cloudinit.go
Diegomangasco: Ok, remove useless declarations

operators/pkg/instctrl/containers.go
Diegomangasco: Ok, removed

operators/pkg/instctrl/containers.go
Diegomangasco: Ok, removed checks for mode and tenant name

operators/pkg/instctrl/controller_test.go
Diegomangasco: Ok, tried to remove it. If it will be required I will add it again

operators/pkg/tenant-controller/tenant_controller.go
Diegomangasco: Ok, for the moment I only changed the variable name from `UserPvcSize` to `UserPvcSizeBytes`

operators/pkg/instctrl/cloudinit.go
```diff
@@ -61,12 +54,20 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
 	}
 	log.V(utils.LogDebugLevel).Info("public keys correctly retrieved")
 
-	userdata, err := forge.CloudInitUserData(r.ServiceUrls.NextcloudBaseURL, user, password, publicKeys)
+	server, path, err := r.GetNFSSpecs(ctx)
+
+	if err != nil {
+		if env.MountPathEnabled {
```
giorio94, line 60: It would be better to avoid retrieving the NFSSpecs altogether if mounting is not enabled.

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,6 +87,33 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
 	return nil
 }
 
+// Get the specs of the volume
```
giorio94, line 90: This comments looks difficult to understand to me, unless one already knows what you are doing.

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,6 +87,33 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
...
+	if err = r.Get(ctx, secretName, &secret); err != nil {
+		ctrl.LoggerFrom(ctx).Error(err, "failed to retrieve secret", "secret", secretName)
+		return
+	}
+
+	if serverNameBytes, ok := secret.Data[NFSSecretServerNameKey]; !ok {
+		err = fmt.Errorf("cannot find %v key in secret", NFSSecretServerNameKey)
+		ctrl.LoggerFrom(ctx).Error(err, "failed to retrieve NFS spec from secret", "secret", secretName)
+		return
+	} else {
```
giorio94, line 105: The else block is not required, since you previously returned in case of error.

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,6 +87,33 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
...
+		if serverPathBytes, ok := secret.Data[NFSSecretPathKey]; !ok {
+			err = fmt.Errorf("cannot find %v key in secret", NFSSecretPathKey)
+			ctrl.LoggerFrom(ctx).Error(err, "failed to retrieve NFS spec from secret", "secret", secretName)
+			return
+		} else {
```
giorio94, line 110: Same here.

operators/pkg/instctrl/cloudinit_test.go
```diff
@@ -116,21 +116,21 @@ var _ = Describe("Generation of the cloud-init configuration", func() {
...
```
giorio94, line 133: ServiceUrls initialization could be dropped.

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,6 +87,33 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
...
+			return string(serverNameBytes), string(serverPathBytes), nil
+		}
+	}
+
+}
```
giorio94, line 115: Feel free to delete webdav related parts from the instance controller, it is probably easier and cleaner.

operators/pkg/instctrl/cloudinit_test.go
```diff
@@ -65,10 +66,9 @@ var _ = Describe("Generation of the cloud-init configuration", func() {
 		environmentName   = "control-plane"
 		tenantName        = "tester"
 
-		nextcloudBaseURL  = "https://nextcloud.example.com"
-		webdavCredentials = "webdav-credentials"
-		webdavUsername    = "username"
-		webdavPassword    = "password"
+		NFSServiceName = "rook-ceph-nfs-my-nfs-a.rook-ceph.svc.cluster.local"
```
giorio94, line 69: Keep test constants simple, they do not need to fully respect reality.

operators/pkg/instctrl/const.go
```diff
@@ -29,4 +29,14 @@ const (
...
+	// PVC name
+	UserPvcName = "user-pvc"
+	// NFS secret name
+	NFSSecretName = "user-pvc-secret"
+	// NFS Server key in NFS secret
+	NFSSecretServerNameKey = "server-name"
+	// NFS path key in NFS secret
+	NFSSecretPathKey = "path"
+	// The path where to mount the NFS volume
+	NFSMountPath = "/media/MyDrive"
```
giorio94, line 41: Constants shall not be duplicated among controllers, otherwise they become useless.

operators/pkg/instctrl/containers.go
```diff
@@ -89,6 +110,22 @@ func (r *InstanceReconciler) enforceContainer(ctx context.Context) error {
...
+					return err
+				}
+
+				depl.Spec.Template.Spec.Volumes = append(depl.Spec.Template.Spec.Volumes, volume)
+				for i := range depl.Spec.Template.Spec.Containers {
+					if depl.Spec.Template.Spec.Containers[i].Name == environment.Name {
+						depl.Spec.Template.Spec.Containers[i].VolumeMounts = append(depl.Spec.Template.Spec.Containers[i].VolumeMounts, volumeMount)
+					}
+				}
+			}
```
giorio94, line 128: Kubernetes objects forging logic shall be in forge, and not here. That logic shall not include calls to the Kubernetes APIs (i.e., it shall never fail), hence the retrieval of the NFS URL and path shall be performed in this method in advance.

operators/pkg/instctrl/suite_test.go
```diff
@@ -58,7 +58,12 @@ var (
 	whiteListMap = map[string]string{"production": "true"}
 )
 
-const webdavSecretName = "webdav-secret"
+const (
+	webdavSecretName       = "webdav-secret"
+	NFSSecretName          = "user-pvc-secret"
+	NFSSecretServerNameKey = "server-name"
+	NFSSecretPathKey       = "path"
+)
```
giorio94, line 66: Constants shall not be duplicated.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -46,6 +47,20 @@ import (
...
+	// NFS Server key in NFS secret
+	NFSSecretServerNameKey = "server-name"
+	// NFS path key in NFS secret
+	NFSSecretPathKey = "path"
+	// PVC size
+	UserPvcSizeBytes = 1 * 1024 * 1024 * 1024 // 1 Gi
+	// Name of the storage class to use
+	StorageClassName = "rook-nfs"
+	// Namespace of rook-ceph resources
+	RookCephNamespace = "rook-ceph"
```
giorio94, line 63: This shall not be hardcoded but parameters.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -395,6 +415,40 @@ func (r *TenantReconciler) createOrUpdateClusterResources(ctx context.Context, t
...
+			})
+			if err != nil {
+				klog.Errorf("Unable to create or update PVC Secret for tenant %s -> %s", tn.Name, err)
+				retErr = err
+			}
+			klog.Infof("PVC Secret for tenant %s %s", tn.Name, pvcSecOpRes)
+		}
+	} else if pvc.Status.Phase == v1.ClaimPending {
+		klog.Infof("PVC pending for tenant %s", tn.Name)
+	}
```
giorio94, line 450: I would personally extract this to a dedicated function.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -395,6 +415,40 @@ func (r *TenantReconciler) createOrUpdateClusterResources(ctx context.Context, t
...
+
+	if pvc.Status.Phase == v1.ClaimBound {
+		pv := v1.PersistentVolume{ObjectMeta: metav1.ObjectMeta{Name: pvc.Spec.VolumeName}}
+		if err := r.Get(ctx, types.NamespacedName{Name: pv.Name}, &pv); err != nil {
+			klog.Errorf("Unable to get PV for tenant %s -> %s", tn.Name, err)
+			retErr = err
+		} else {
+			pvcSecret := v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: NFSSecretName, Namespace: nsName}}
+			pvcSecOpRes, err := ctrl.CreateOrUpdate(ctx, r.Client, &pvcSecret, func() error {
+				r.updateTnPVCSecret(&pvcSecret, fmt.Sprintf("%s.%s.svc.cluster.local", pv.Spec.CSI.VolumeAttributes["server"], RookCephNamespace), pv.Spec.CSI.VolumeAttributes["share"])
```
giorio94, line 439: Do not hard-code svc.cluster.local, it is not needed.

operators/pkg/tenant-controller/tenant_controller_test.go
```diff
@@ -213,11 +213,51 @@ var _ = Describe("Tenant controller", func() {
...
+						},
+					},
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, pv)).Should(Succeed())
+		pvc.Spec.VolumeName = "pv-test"
+		Expect(k8sClient.Update(ctx, &pvc)).Should(Succeed())
+		pvc.Status.Phase = v1.ClaimBound
+		Expect(k8sClient.Status().Update(ctx, &pvc)).Should(Succeed())
```
giorio94, line 256: Should be moved within/after `checkTnClusterResourceCreation`, possibly in a dedicated function.

giorio94 (commented):
Overall this looks pretty good to me. I left you a few comments inline. 
Please rebase also your PR onto master, and squash all commits.

operators/api/v1alpha2/template_types.go
```diff
@@ -122,6 +122,10 @@ type Environment struct {
 
 	// Name of the storage class to be used for the persistent volume (when needed)
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// +kubebuilder:default=true
+	// Whether the instance has to have the user path
```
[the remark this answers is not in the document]
matteorosani: Something like `MountNFSPersonalVolume` could be fine?

operators/pkg/forge/cloudinit.go
```diff
@@ -58,7 +57,7 @@ type writefile struct {
 }
 
 // CloudInitUserData forges the yaml manifest representing the cloud-init userdata configuration.
-func CloudInitUserData(nextcloudBaseURL, webdavUsername, webdavPassword string, publicKeys []string) ([]byte, error) {
+func CloudInitUserData(serviceName string, servicePath string, publicKeys []string) ([]byte, error) {
```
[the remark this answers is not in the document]
matteorosani: Changed to `NFSServerName`

operators/pkg/instctrl/cloudinit.go
matteorosani: Fixed

operators/pkg/instctrl/cloudinit.go
matteorosani: Added clearer comment

operators/pkg/instctrl/cloudinit.go
matteorosani: Removed else block

operators/pkg/tenant-controller/tenant_controller.go
matteorosani: We should get them from the tenant controller's arguments?

operators/pkg/instctrl/suite_test.go
matteorosani: Removed duplicated constants, now they are defined only in tenant_controller

operators/pkg/tenant-controller/tenant_controller.go
matteorosani: Removed. But the rook ceph namespace is needed right?

operators/pkg/tenant-controller/tenant_controller.go
matteorosani: Moved

operators/api/v1alpha2/template_types.go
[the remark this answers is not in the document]
giorio94: I would propose just MountPersonalVolume, to make it unrelated from the current implementation.

operators/pkg/tenant-controller/tenant_controller.go
giorio94: Yes, exactly

operators/pkg/tenant-controller/tenant_controller.go
giorio94: Yup!

operators/api/v1alpha2/template_types.go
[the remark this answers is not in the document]
matteorosani: Ok

operators/pkg/tenant-controller/tenant_controller.go
matteorosani: Ok

operators/pkg/tenant-controller/tenant_controller_test.go
matteorosani: Moved

operators/pkg/instctrl/cloudinit_test.go
matteorosani: Removed

operators/pkg/instctrl/cloudinit_test.go
matteorosani: Simplified

operators/pkg/instctrl/const.go
matteorosani: Removed unnecessary ones

operators/pkg/instctrl/containers.go
matteorosani: Thinking of a way of moving it

matteorosani: A possible solution has been provided, take a look at it. I don't know if the approach is completely correct.

operators/pkg/tenant-controller/tenant_controller.go
Diegomangasco: Should we do some type of controls or checks for these arguments?

Diegomangasco: Added tenant controller arguments for PVC parameters

operators/pkg/instctrl/containers.go
```diff
@@ -28,6 +28,27 @@ import (
 	"github.com/netgroup-polito/CrownLabs/operators/pkg/utils"
 )
 
+const (
+	NFSVolumeName = "nfs-vol"
+)
+
+func (r *InstanceReconciler) GetNFSVolume(ctx context.Context, volume *v1.Volume, volumeMount *v1.VolumeMount) error {
```
giorio94, line 35: Seems unused to me, and could be dropped.

operators/pkg/instctrl/suite_test.go
```diff
@@ -55,7 +55,9 @@ var (
 	whiteListMap = map[string]string{"production": "true"}
 )
 
-const webdavSecretName = "webdav-secret"
+const (
+	webdavSecretName = "webdav-secret"
+)
```
giorio94, line 60: Could be probably dropped.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -396,9 +413,49 @@ func (r *TenantReconciler) createOrUpdateClusterResources(ctx context.Context, t
 		retErr = err
 	}
 	klog.Infof("Allow network policy for tenant %s %s", tn.Name, npAOpRes)
+
+	retErr = r.createOrUpdateTnPersonalNFSVolume(ctx, tn, nsName)
```
giorio94, line 417: To keep the previous logic, you should not override `retErr` in case it is nil.

operators/pkg/tenant-controller/tenant_controller_test.go
```diff
@@ -211,11 +211,12 @@ var _ = Describe("Tenant controller", func() {
 			},
 		}
 		Expect(k8sClient.Create(ctx, tn)).Should(Succeed())
+		tnNm := &v1.Namespace{}
+		doesEventuallyExists(ctx, types.NamespacedName{Name: nsName}, tnNm, BeTrue(), timeout, interval)
```
giorio94, line 215: This is already checked in `checkTnClusterResourceCreation`.

giorio94 (commented):
A few more comments inline. Please, squash also all commits into one.

operators/pkg/forge/cloudinit.go
```diff
@@ -58,7 +57,7 @@ type writefile struct {
 }
 
 // CloudInitUserData forges the yaml manifest representing the cloud-init userdata configuration.
-func CloudInitUserData(nextcloudBaseURL, webdavUsername, webdavPassword string, publicKeys []string) ([]byte, error) {
+func CloudInitUserData(NFSServerName string, NFSPath string, publicKeys []string) ([]byte, error) {
```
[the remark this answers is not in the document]
matteorosani: But NFS is an acronym, is it fine if I call it nfs?

operators/cmd/tenant-operator/main.go
```diff
@@ -169,6 +176,16 @@ func main() {
...
 	}
+
+	env := clctx.EnvironmentFrom(ctx)
+	if env.MountPersonalVolume && (userPvcSizeBytes == 0 || storageClassName == "" || rookCephNamespace == "") {
+		log.Error(errors.New("missing parameters for user's personal volume"), "Initialization failed")
+		os.Exit(1)
+	} else if !env.MountPersonalVolume && (userPvcSizeBytes != 0 || storageClassName != "" || rookCephNamespace != "") {
+		log.Error(errors.New("the environment is not set to enable the user's personal volume"), "Initialization failed")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
matteorosani: Removed

operators/pkg/instctrl/containers.go
```diff
@@ -28,6 +28,27 @@ import (
 	"github.com/netgroup-polito/CrownLabs/operators/pkg/utils"
 )
 
+const (
+	NFSVolumeName = "nfs-vol"
+)
```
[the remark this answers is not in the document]
matteorosani: Not used

operators/pkg/instctrl/containers.go
```diff
@@ -28,6 +28,27 @@ import (
...
+
+func (r *InstanceReconciler) GetNFSVolume(ctx context.Context, volume *v1.Volume, volumeMount *v1.VolumeMount) error {
+
+	server, path, err := r.GetNFSSpecs(ctx)
+	if err != nil {
+		klog.Errorf("Unable to get secret for user pvc in namespace. Error %s", err)
+		return err
+	}
+	klog.Infof("Adding NFS volume %s:%s", string(server), string(path))
```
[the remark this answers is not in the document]
matteorosani: We should use the logger from context?

operators/pkg/forge/containers.go
```diff
@@ -467,6 +451,20 @@ func ContainerVolume(volumeName, claimName string, environment *clv1alpha2.Envir
 	}
 }
 
+// TenantPersonalNFSVolume returns the Personal volume of the tenant
+func TenantPersonalNFSVolume(volumeName, NFSServerName, NFSPath string) corev1.Volume {
```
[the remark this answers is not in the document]
matteorosani: Moved

operators/pkg/instctrl/containers.go
matteorosani: Dropped

operators/pkg/forge/cloudinit.go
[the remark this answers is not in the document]
giorio94: Yea, no issue at all, as long as it is consistent (either all lowercase or uppercase).

operators/pkg/instctrl/containers.go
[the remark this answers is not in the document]
giorio94: Yes, exactly, as already done in the other functions.

operators/pkg/instctrl/const.go
```diff
@@ -29,4 +29,6 @@ const (
 	EvEnvironmentErr = "EnvironmentEnforcementFailed"
 	// EvEnvironmentErrMsg -> the event message corresponding to a failed environment enforcement.
 	EvEnvironmentErrMsg = "Failed to enforce environment %v"
+	// The path where to mount the NFS volume
+	NFSMountPath = "/media/MyDrive"
```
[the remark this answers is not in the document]
matteorosani: I don't understand what I have to do

operators/pkg/forge/cloudinit.go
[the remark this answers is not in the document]
matteorosani: But everywhere else when NFS is present is all capital

operators/pkg/tenant-controller/tenant_controller.go
matteorosani: Ok

operators/pkg/forge/cloudinit.go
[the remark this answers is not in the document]
giorio94: I mean for a single variable name. That is, not something like NfsServerName.

operators/pkg/tenant-controller/tenant_controller_test.go
matteorosani: Removed

operators/cmd/tenant-operator/main.go
```diff
@@ -92,6 +96,9 @@ func main() {
 	flag.StringVar(&webhookBypassGroups, "webhook-bypass-groups", "system:masters", "The list of groups which can skip webhooks checks, comma separated values")
 	sandboxClusterRole := flag.String("sandbox-cluster-role", "crownlabs-sandbox", "The cluster role defining the permissions for the sandbox namespace.")
 	enableWH := flag.Bool("enable-webhooks", true, "Enable webhooks server")
+	flag.IntVar(&userPvcSizeBytes, "user-pvc-size-bytes", 1*1024*1024*1024, "The dimension of the user's personal space")
```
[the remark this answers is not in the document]
matteorosani: Using the custom flag type how can I set the default value?

operators/pkg/instctrl/suite_test.go
matteorosani: Removed

operators/pkg/instctrl/const.go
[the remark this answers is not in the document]
Diegomangasco: So /media/mydrive should be the not persistent and /media/MyDrive (or another name as you want) should be the persistent, right?

operators/cmd/tenant-operator/main.go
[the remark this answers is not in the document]
giorio94: It is enough to initialize the custom arg with that value.

operators/pkg/instctrl/const.go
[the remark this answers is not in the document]
giorio94: No, /media/mydrive should be the one bound to the NFS personal volume, while another path should be the one mounted to the per instance PVC if present (the one currently called mydrive).

operators/cmd/tenant-operator/main.go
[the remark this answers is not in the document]
matteorosani: Ok

operators/pkg/forge/containers_test.go
```diff
@@ -77,9 +77,10 @@ var _ = Describe("Containers and Deployment spec forging", func() {
 			ObjectMeta: metav1.ObjectMeta{Name: instanceName, Namespace: instanceNamespace},
 		}
 		environment = clv1alpha2.Environment{
-			Image: image,
-			Name:  envName,
-			Mode:  clv1alpha2.ModeStandard,
+			Image:               image,
+			Name:                envName,
+			Mode:                clv1alpha2.ModeStandard,
+			MountPersonalVolume: true,
```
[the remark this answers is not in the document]
matteorosani: I've added a check in the `The forge.ContainerVolumes function` describe block

operators/pkg/instctrl/const.go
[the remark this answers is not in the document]
matteorosani: I think I've done it correctly, take a look

operators/pkg/tenant-controller/tenant_controller_test.go
```diff
@@ -398,4 +439,20 @@ func checkTnClusterResourceCreation(ctx context.Context, tnName, nsName string,
...
+	doesEventuallyExists(ctx, NFSSecretLookupKey, createdNFSSecret, BeTrue(), timeout, interval)
+
+	By("By checking that the NFS pvc secret of the tenant has a owner reference and label pointing to the tenant operator")
+	Expect(createdNFSSecret.OwnerReferences).Should(ContainElement(MatchFields(IgnoreExtras, Fields{"Name": Equal(tnName)})))
+	Expect(createdNFSSecret.Labels).Should(HaveKeyWithValue("crownlabs.polito.it/managed-by", "tenant"))
+
+	By("By checking that the NFS pvc secret of the tenant has a correct spec")
+	Expect(createdNFSSecret.Type).Should(Equal(v1.SecretTypeOpaque))
+	Expect(createdNFSSecret.Data).Should(HaveKey(NFSSecretServerNameKey))
+	Expect(createdNFSSecret.Data).Should(HaveKey(NFSSecretPathKey))
```
giorio94, line 457: Nit. You could check that also the value is correct.

operators/pkg/forge/containers.go
```diff
@@ -61,6 +61,10 @@ const (
 	AppMEMLimitsEnvName = "APP_MEM_LIMITS"
 	// PodNameEnvName -> name of the env variable containing the Pod Name.
 	PodNameEnvName = "POD_NAME"
+	// Name of the NFS volume
+	NFSVolumeName = "nfs-vol"
+	// Mount path for the NFS personal volume
+	NFSVolumeMountPath = "/media/MyDrive"
```
giorio94, line 67: Still some issues with naming here. Having `/mydrive` and `media/MyDrive` is likely to lead to confusion. To me, the feature you are introducing is the one that should be named mydrive (for consistency with the VMs), while the previous one should be changed to something else (e.g., `/media/persistent`). Also the variable names should reflect this. @QcFe WDYT?

operators/pkg/instctrl/cloudinit.go
```diff
@@ -86,35 +90,34 @@ func (r *InstanceReconciler) EnforceCloudInitSecret(ctx context.Context) error {
 	return nil
 }
 
-// GetWebDavCredentials extracts the credentials (i.e. username and password)
+// Extracts the NFS server name and path for the user's personal NFS volume,
```
giorio94, line 93: The comment should start with the name of the function.

operators/pkg/instctrl/containers_test.go
```diff
@@ -102,17 +106,28 @@ var _ = Describe("Generation of the container based instances", func() {
...
 
+		tenant = clv1alpha2.Tenant{
+			ObjectMeta: metav1.ObjectMeta{Name: tenantName},
+			Status: clv1alpha2.TenantStatus{
+				PersonalNamespace: clv1alpha2.NameCreated{
+					Created: true,
+					Name:    instanceNamespace,
+				},
+			},
+		}
```
giorio94, line 117: Is this required?

operators/pkg/instctrl/controller_test.go
```diff
@@ -82,15 +83,23 @@ var _ = Describe("The instance-controller Reconcile method", func() {
...
```
giorio94, line 102: Should be removed.

giorio94 (commented):
Almost ready to test this feature in the cluster... Please, squash also the commits.

operators/pkg/instctrl/controller_test.go
Diegomangasco: Removed

operators/pkg/instctrl/cloudinit.go
Diegomangasco: Changed

operators/pkg/forge/containers.go
Diegomangasco: Names changed

operators/pkg/instctrl/containers_test.go
Diegomangasco: Ok removed tenant

operators/pkg/tenant-controller/tenant_controller_test.go
matteorosani: Done

operators/pkg/forge/containers.go
```diff
@@ -61,6 +61,10 @@ const (
 	AppMEMLimitsEnvName = "APP_MEM_LIMITS"
 	// PodNameEnvName -> name of the env variable containing the Pod Name.
 	PodNameEnvName = "POD_NAME"
+	// Name of the NFS volume
+	NFSVolumeName = "nfs-vol"
+	// Mount path for the NFS personal volume
+	NFSVolumeMountPath = "/mydrive"
```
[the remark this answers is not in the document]
Diegomangasco: Ok I removed `MyDriveDBPath`, and change the variable names from `NFS[Something]` to `MyDrive[Something]`.
Now the VMs have the path `/media/mydrive`, and the Containers have the paths `/media/mydrive` and `/media/persistent`.
In that way `/media/mydrive` is the NFS storage both for VMs and Containers, and for Containers we have the previous one in `/media/persistent`.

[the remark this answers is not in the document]
Diegomangasco: And also the `MyDriveName` for FileBrowser now is `FileBrowserName`.

operators/pkg/forge/containers.go
```diff
@@ -61,6 +59,10 @@ const (
 	AppMEMLimitsEnvName = "APP_MEM_LIMITS"
 	// PodNameEnvName -> name of the env variable containing the Pod Name.
 	PodNameEnvName = "POD_NAME"
+	// Name of the NFS volume
+	MyDriveVolumeName = "nfs-vol"
```
giorio94, line 63:
```suggestion
	MyDriveVolumeName = "mydrive"
```

operators/pkg/forge/containers.go
```diff
@@ -37,16 +37,14 @@ const (
 	WebsockifyName = "websockify"
 	// XVncName -> name of the x+vnc server sidecar container.
 	XVncName = "xvnc"
-	// MyDriveName -> name of the filebrowser sidecar container.
-	MyDriveName = "mydrive"
+	// FileBrowserName -> name of the filebrowser sidecar container.
+	FileBrowserName = "mydrive"
```
giorio94, line 41:
```suggestion
	// PersistentVolumeName -> name of the persistent volume.
	PersistentVolumeName = "persistent"
```

operators/pkg/instmetrics/metrics_scraper.go
```diff
@@ -84,7 +84,7 @@ func (ms *MetricsScraper) Start(ctx context.Context) {
 		forge.XVncName,
 		forge.WebsockifyName,
 		forge.ContentDownloaderName,
-		forge.MyDriveName,
+		forge.FileBrowserName,
```
giorio94, line 87: Could be dropped, since you removed that container.

giorio94 (commented):
A couple of nits. Commits should also be squashed again.

operators/pkg/forge/containers.go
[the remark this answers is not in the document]
giorio94: :+1:

operators/pkg/forge/containers.go
Diegomangasco: Done

operators/pkg/forge/containers.go
Diegomangasco: Done

operators/pkg/instmetrics/metrics_scraper.go
Diegomangasco: Done

Diegomangasco:
Ok now we will squash commits

operators/pkg/forge/cloudinit.go
```diff
@@ -51,14 +50,14 @@ type interf struct {
...
-type writefile struct {
-	Content     string `yaml:"content"`
-	Path        string `yaml:"path"`
-	Permissions string `yaml:"permissions"`
-}
+// type writefile struct {
+// 	Content     string `yaml:"content"`
+// 	Path        string `yaml:"path"`
+// 	Permissions string `yaml:"permissions"`
+// }
```
giorio94, line 57: Should be dropped.

operators/pkg/forge/containers_test.go
```diff
@@ -64,12 +64,9 @@ var _ = Describe("Containers and Deployment spec forging", func() {
...
 		httpPathAlternative  = "/some/different/path"
-	)
-
-	// test aware constants (are tested against code)
-	const (
-		myDriveMountPath = "/mydrive"
-		myDriveName      = "mydrive"
+		nfsServerName        = "nfs-server-name"
+		nfsPath              = "/nfs/path"
+		nfsVolumeName        = "nfs-vol"
```
giorio94, line 69: This constant is unused, and should be dropped.

operators/pkg/forge/containers.go
```diff
@@ -467,6 +458,20 @@ func ContainerVolume(volumeName, claimName string, environment *clv1alpha2.Envir
 	}
 }
 
+// TenantPersonalNFSVolume returns the Personal volume of the tenant
+func TenantPersonalNFSVolume(volumeName, nfsServerName, nfsPath string) corev1.Volume {
```
giorio94, line 462:
```suggestion
func MyDriveVolume(volumeName, nfsServerName, nfsPath string) corev1.Volume {
```

A bit more consistent IMHO.

operators/api/v1alpha2/template_types.go
```diff
@@ -122,6 +122,10 @@ type Environment struct {
 
 	// Name of the storage class to be used for the persistent volume (when needed)
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// +kubebuilder:default=true
+	// Whether the instance has to have the user's personal NFS volume
+	MountPersonalVolume bool `json:"mountPersonalVolume"`
```
giorio94, line 128:
```suggestion
	MountMyDriveVolume bool `json:"mountMyDriveVolume"`
```

Wondering whether this naming would be clearer. @QcFe WDYT?

operators/pkg/instctrl/containers_test.go
```diff
@@ -107,12 +110,14 @@ var _ = Describe("Generation of the container based instances", func() {
...
 			},
 		}
 		environment = clv1alpha2.Environment{
-			Name:            environmentName,
-			EnvironmentType: clv1alpha2.ClassContainer,
-			Image:           image,
+			Name:                environmentName,
+			EnvironmentType:     clv1alpha2.ClassContainer,
+			Image:               image,
+			MountPersonalVolume: false,
```
giorio94, line 120: Could set to true (and set valid values for the server and the path), to also test your feature.

giorio94 (commented):
A couple of very minor comments inline. There are also still a few issues with the linter, and the automatic generation of manifests to be fixed.

operators/pkg/forge/cloudinit.go
Diegomangasco: Removed

operators/pkg/forge/containers_test.go
Diegomangasco: Removed

operators/pkg/forge/containers.go
Diegomangasco: Changed

operators/api/v1alpha2/template_types.go
Diegomangasco: Changed

operators/pkg/instctrl/containers_test.go
Diegomangasco: Change to true and assign valid names to server and path

operators/api/v1alpha2/template_types.go
```diff
@@ -122,6 +122,10 @@ type Environment struct {
 
 	// Name of the storage class to be used for the persistent volume (when needed)
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// +kubebuilder:default=true
+	// Whether the instance has to have the user's personal NFS volume
+	MountMyDriveVolume bool `json:"mountPersonalVolume"`
```
giorio94, line 128:
```suggestion
	MountMyDriveVolume bool `json:"mountMyDriveVolume"`
```

operators/deploy/tenant-operator/templates/deployment.yaml
```diff
@@ -47,6 +47,10 @@ spec:
             - "--webhook-bypass-groups={{ .Values.webhook.deployment.webhookBypassGroups }}"
             - "--sandbox-cluster-role={{ .Values.configurations.sandboxClusterRole }}"
             - "--max-concurrent-reconciles={{ .Values.configurations.maxConcurrentReconciles }}"
+            - "--user-pvc-size-bytes={{ .Values.configurations.userPvcSizeBytes }}"
+            - "--storage-class-name={{ .Values.configurations.storageClassName }}"
+            - "--rook-ceph-namespace={{ .Values.configurations.rookCephNamespace }}"
+            - "--mydrive-pvcs-namespace={{ .Values.configurations.mydrivePvcsNamespace }}"
```
giorio94, line 53: I would uniform the names of this parameters (both the args and in Helm), so that they all start with mydrive-pvcs-.

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -398,9 +414,54 @@ func (r *TenantReconciler) createOrUpdateClusterResources(ctx context.Context, t
...
+
+	if pvc.Status.Phase == v1.ClaimBound {
+		pv := v1.PersistentVolume{ObjectMeta: metav1.ObjectMeta{Name: pvc.Spec.VolumeName}}
+		if err := r.Get(ctx, types.NamespacedName{Name: pv.Name}, &pv); err != nil {
+			klog.Errorf("Unable to get PV for tenant %s -> %s", tn.Name, err)
+			return err
+		}
+		pvcSecret := v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: NFSSecretName, Namespace: nsName}}
+		pvcSecOpRes, err := ctrl.CreateOrUpdate(ctx, r.Client, &pvcSecret, func() error {
+			r.updateTnPVCSecret(&pvcSecret, fmt.Sprintf("%s.%s", pv.Spec.CSI.VolumeAttributes["server"], r.RookCephNamespace), pv.Spec.CSI.VolumeAttributes["share"])
```
giorio94, line 451: You could drop the RookCephNamespace parameter, and take that value from the `clusterID` volume attribute (which corresponds to the namespace).

operators/deploy/tenant-operator/values.yaml
```diff
@@ -17,6 +17,10 @@ configurations:
     url: "https://nextcloud.crownlabs.example.com/"
     user: username
     pass: password
+  userPvcSizeBytes: 1Gi
```
giorio94, line 20:
```suggestion
  userPvcSize: 1Gi
```

operators/deploy/tenant-operator/values.yaml
```diff
@@ -17,6 +17,10 @@ configurations:
...
+  storageClassName: rook-nfs
+  rookCephNamespace: rook-ceph
+  mydrivePvcsNamespace: mydrive-pvcs
```
giorio94, line 23: Same here about starting these variables with mydrivePVCs.

giorio94 (approved):
Just very few more nits, then you should also propagate the values file modifications to the main file in deploy/crownlabs.

operators/deploy/tenant-operator/values.yaml
matteorosani: For this and the other comments, do you want `mydrivePVCs...` or `mydrivePvcs...`?

giorio94: To me mydrivePVCs... is slightly better, but that's not a big deal.

operators/deploy/tenant-operator/values.yaml
matteorosani: To uniform it to the other parameter's name is it ok to call it `mydrivePVCsSize`?

giorio94: Absolutely

QcFe (approved):

PR #833, opened 2023-01-10
Remove nextcloud

matteorosani:
# Description

In this pull request we aim at removing the NextCloud service because not used anymore after the introduction of personal NFS storage.


frisso:
@QcFe How do we handle this PR?
If we merge the PR, we should also update all the images for VMs and containers, and update the dashboard, isn't it?

QcFe:
With the previous PR we already are in an unstable/undeployable configuration. We'll need at least one more PR for the frontend and indeed to rebuild all the VM images. Containers need no rebuild since they didn't use NextCloud.

frisso:
Thanks, @QcFe . Please go the way you think is the best.

QcFe:
@matteorosani reporting here too, just to keep track, found some more stuff to be cleaned:
- in `deploy/crownlabs/values.yaml` there are still some NC parameters
- remove `infrastructure/file-sharing`
- some params quoted still in `operators/README.md`

QcFe:
@cannarelladev would you mind doing a check on the frontend and removing the myDrive button and references? I fear we'll need to wait to deploy the new CRDs (without the myDrive) in order to have the typescript generator thingy to forget about it, but if we start to remove references it should break a bit less once we remove them...

QcFe (approved):

PR #850, opened 2023-05-10
Harbor documentation: fix dead link

air-31:
# Description

This PR fixes a dead link present in the Harbor (docker registry) documentation.


cheina97 (approved):

frisso (approved):
Is there any reasons we're using link to version 2.7.0, which is not thd latest?

air-31:
Yes, the reason is the absence of the corresponding page in the latest version documentation

QcFe:
@frisso guess when I updated it that was the latest available version...
It seems the 2.8 release has been published just 3 weeks ago :)

PR #875, opened 2023-11-04
Removed quotas cap

cheina97:
# Description

This PR removes the cap on CPU and memory for tenants.

@QcFe can you confirm if it is secure? If I remember well a normal user cannot customize it's tenant.

### DIFF
--- a/operators/deploy/tenant-operator/values.yaml
+++ b/operators/deploy/tenant-operator/values.yaml
@@ -19,6 +19,14 @@ configurations:
   maxConcurrentReconciles: 1
   sandboxClusterRole: crownlabs-sandbox
   tenantNamespaceKeepAlive: 168h
+  tenant:
+    resourcecaps:
+      # Maximum number of virtual cpus that can be requested by a tenant. 0 means unlimited.
+      cpu: 25
+      # Maximum amount of memory that can be requested by a tenant (expressed in GB). 0 means unlimited.
+      memory: 50
+      # Maximum number of instances that can be created by a tenant. 0 means unlimited.
+      instances: 10
 
 image:
   repository: crownlabs/tenant-operator
--- a/operators/pkg/forge/resourcequota.go
+++ b/operators/pkg/forge/resourcequota.go
@@ -25,17 +25,17 @@ import (
 const (
 	// InstancesCountKey -> The key for accessing at the total number of instances in the corev1.ResourceList map.
 	InstancesCountKey = "count/instances.crownlabs.polito.it"
-
-	// CapInstance -> The cap number of instances that can be requested by a Tenant.
-	CapInstance = 10
 )
 
 var (
-	// CapCPU -> The cap amount of CPU cores that can be requested by a Tenant.
-	CapCPU = *resource.NewQuantity(25, resource.DecimalSI)
+	// CapInstance -> The maximum number of instances that can be started by a Tenant.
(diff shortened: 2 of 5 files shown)


frisso:
Wouldn't be possible to set those quota parameters as properties of the tenant itself? With "zero" as "no limit"?

cheina97:
> Wouldn't be possible to set those quota parameters as properties of the tenant itself? With "zero" as "no limit"?

They should be added in the tenant spec. This means a change in the Crownlabs API. I think it should be a good improvement, but it could require some times.

cheina97:
> Would you mind setting the new values inside the helm chart too? Setting as default the old values that were hardcoded...

Of course 👍

cheina97:
@QcFe PR is ready or review

operators/deploy/tenant-operator/values.yaml
```diff
@@ -19,6 +19,14 @@ configurations:
   maxConcurrentReconciles: 1
   sandboxClusterRole: crownlabs-sandbox
   tenantNamespaceKeepAlive: 168h
+  tenant:
+    resourcecaps:
+      # Maximum number of virtual cpus that can be requested by a tenant.
```
frisso, line 24: I would write here whether those limits are *per instance* or *per tenant*.
If it is *per tenant*, those numbers look pretty small.

frisso:
If we want to give resources to our students that exceed their capabilities at home, I feel that, per each instance, students should be allowed to create up to 4 cores and 16GB RAM at least (personally, I would say even 8 and 32).
Max 10 instances, instead, looks fair to me.
@QcFe @cheina97 what do you think?

operators/deploy/tenant-operator/values.yaml
cheina97: I think @QcFe is going to override those values with ArgoCD or through CrownOps. But we can surely increase the default value. I forgot to add in the description that 0 means that the it no cap in resources, I'm going to add it.

operators/pkg/forge/resourcequota.go
```diff
@@ -25,17 +25,17 @@ import (
...
-
-	// CapInstance -> The cap number of instances that can be requested by a Tenant.
-	CapInstance = 10
 )
 
 var (
+	// CapInstance -> The cap number of instances that can be requested by a Tenant.
+	CapInstance int
+
 	// CapCPU -> The cap amount of CPU cores that can be requested by a Tenant.
```
frisso, line 34:
```suggestion
	// CapCPU -> The total amount of CPU cores that can be requested by a Tenant.
```

frisso:
> > If we want to give resources to our students that exceed their capabilities at home, I feel that, per each instance, students should be allowed to create up to 4 cores and 16GB RAM at least (personally, I would say even 8 and 32). Max 10 instances, instead, looks fair to me. @QcFe @cheina97 what do you think?
> 
> At the moment the caps are 25 for the CPU, 50GB for RAM and 10 for the instances. What do you think about putting a huge amount of resources as cap? Something like 150 CPU, 300GB of RAM, and 50 instances? Remember that this limit is for all tenants (not only students). The limits of student's resources should be managed with workspaces.

Current limits (25 CPU, 50GB, 10 instances) look perfect to me.

frisso:
LGTM, for me we can merge.

PR #914, opened 2024-11-15
Fix SSH key deletion error bug

Mattia013:
# Description

Fix delete modal to prompt for the specific SSH key record in table.

When multiple SSH keys are available and we try to remove one, it is always prompted to remove the last one.
This should fix this undesired behavior.

# How Has This Been Tested?

The test was done recreating the SSHKeysTable in a new React Typescript project.


frisso:
Thanks @Mattia013 for this contribution!
@QcFe This looks very good to me.
If you agree, we can approve and merge.
Thanks to you both!

QcFe (approved):
lgtm, thanks!

PR #916, opened 2024-12-03
Add PVC Provisioning Job

AleTopp:
# Description

Added Job that changes the owner of the MyDrive from root (default) to user 1010 (`CrownLabsUserID`), so that it is writable by tenants inside Instances deployed as containers (that run as user `CrownLabsUserID`). If the Job completes, it adds a label to the PVC `pvc-provisioning: completed` to flag it.

Fixes issue: NFS MyDrive cannot be written from containers since owner is root

~ _Alessio Giliberti, Chiara Mercurio_

### DIFF
--- a/operators/pkg/tenant-controller/tenant_controller.go
+++ b/operators/pkg/tenant-controller/tenant_controller.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	netv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -30,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


QcFe:
Managed to have a look. Overall logic now seems fine. Minor changes:
- switch from CreateOrUpdate to r.Update for pvc labels, since for sure you already have the pvc there.
- change 'pvc-provisioning' to 'mydrive-provisioning'. Also I'd likely add a new 'pending' value that can be used to retrieve all tenants that for any reason are stuck on provisioning.
- as you can see [here](https://github.com/netgroup-polito/CrownLabs/actions/runs/12313295293/job/34489426118?pr=916) there are issues with linting. you can go to your go settings in vscode and select golangci-lint as the linter, to get in sync with the linter we're using in the pipeline
- we'll still deploy now and, if it works, we can still already test the logic

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -53,6 +55,18 @@ const (
 	NFSSecretServerNameKey = "server-name"
 	// NFSSecretPathKey -> NFS path key in NFS secret.
 	NFSSecretPathKey = "path"
+	// ProvisionJobBaseImage -> Base container image for Personal Drive provision job.
+	ProvisionJobBaseImage = "busybox"
+	// ProvisionJobLabel -> Key of the label added by the Provision Job to flag the PVC.
+	ProvisionJobLabel = "mydrive-provisioning"
```
QcFe, line 61: please move labels where others are

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -517,6 +532,32 @@ func (r *TenantReconciler) createOrUpdateTnPersonalNFSVolume(ctx context.Context
 			return err
 		}
 		klog.Infof("PVC Secret for tenant %s %s", tn.Name, pvcSecOpRes)
+
+		val, ok := pvc.Labels[ProvisionJobLabel]
+		if !ok || val != ProvisionJobValueOk {
```
QcFe, line 537: split the logic based on check. if no labels create, if label is not ok, check if it's done, etc

operators/pkg/tenant-controller/tenant_controller.go
```diff
@@ -517,6 +532,32 @@ func (r *TenantReconciler) createOrUpdateTnPersonalNFSVolume(ctx context.Context
...
+
+			if chownJob.Status.Succeeded == 1 {
+				pvc.Labels[ProvisionJobLabel] = ProvisionJobValueOk
+				if err := r.Update(ctx, &pvc); err != nil {
+					klog.Errorf("PVC Provisioning Job failed to update PVC labels for tenant %s", tn.Name)
+				}
+
+				klog.Infof("PVC Provisioning Job completed for tenant %s", tn.Name)
+			} else if chownJob.Status.Failed == 1 {
+				klog.Errorf("PVC Provisioning Job failed for tenant %s", tn.Name)
```
QcFe, line 558: probably better to move into a warning (if exists)
```suggestion
				klog.Warningf("PVC Provisioning Job failed for tenant %s", tn.Name)
```

QcFe (commented):
almost there :)

QcFe (approved):
lgtm

PR #917, opened 2024-12-20
Cloud Image Registry Implementation

air-31:
# Description

This PR implements a  new component - Cloud Image Registry - created with the purpose to store virtual machines disk files inside a storage within the cluster in order to accelerate persistent template instances creation process, and to avoid using kubevirt containerdisks.


QcFe:
Thanks for the latest changes @air-31!
Have a look at the checks to fix the little issues that arose from the renaming to ciregistry, then,
would you mind pull-rebasing from main and simplifying the openapi method names? (removing the "handle" prefix basically)
Eventually please squash the whole thing in a single commit :)

frisso (changes requested):
@air-31 This looks very good to me.
I would kindly ask you to add a page in the documentation (e.g., https://github.com/netgroup-polito/CrownLabs/tree/master/infrastructure) to briefly describe what this component is about, and the minimal info required to use/configure it (info about extending/updating would be nice as well, but not mandatory)

PR #918, opened 2025-01-08
Update years in license headers

air-31:
This PR updates all license headers in Go files inside the operators directory.

The modifications were done with the help of a custom script since the addlicense tool, which was made use of in the Makefile, wasn't working as expected. The update includes files starting with 'zz_generated', since they were not ignored in the Makefile command.


QcFe (approved):
thanks!

PR #943, opened 2025-06-10
3. Clusters-in-clusters using Cluster API  in CrownLabs

KUOHANTUNG:
PR from group 3 Clusters-in-clusters using Cluster API  in CrownLabs

PR open just to show our codes and Implemented functions whether is right. Have tested in local environment, including creating cluster, cni and exposed port.

The exposed way using nginx passthrough needs  target port and nodeport and then certSAN certifacated just by apiserver.

It will first wait the cluster ready then export the kubeconfig file into folders then create cni. but i substitute the service address with external URL name in kubeconfig file, so need to build nginx first.


KUOHANTUNG:
someting i have modifed, e.g add finalizer that it will delete kubeconfigs when delete the instance, adding visualizer which can monitor the  status of each worker

KUOHANTUNG:
The gui code we put in operators/cluster-api-visualizer

PR #948, opened 2025-07-03
Add format on save

umberto-fontanazza:
Not tested

# Description

Add format on save as default config

# How Has This Been Tested?

NOT TESTED

### DIFF
--- a/provisioning/standalone/vscode/rust/settings.json
+++ b/provisioning/standalone/vscode/rust/settings.json
@@ -8,5 +8,9 @@
     "code-runner.executorMap": {
         "rust": "cd $dir.. && cargo run"
     },
-    "rust-analyzer.cache.warmup": false
+    "rust-analyzer.cache.warmup": false,
+    "editor.codeActionsOnSave": {
+      "source.organizeImports": "always"
+    },
+    "editor.formatOnSave": true,
 }


frisso:
@AleTopp (and Chiara... I cannot find your github account) Are you ok with this PR? If so, we can merge.

AleTopp:
We already talked to Federico about it, but didn't come to a conclusion because we couldn't agree on the autosave settings that is linked with this. 
Thanks for bringing this up, we messaged Federico now, and hopefully will come to a conclusion soon.

AleTopp:
Okay, so we tested this with our configuration of autosave (afterDelay of 1000ms), and the code will be formatted only after explicitly saving (with Ctrl+S or the Save button), so apparently there should be no consequences on the load side (that we were worried about).

But since the auto-format is just a decorative setting, that could be activated in any moment by anyone, we don't feel like this should be the default.
We also asked to professor Rebaudengo, since this image is for his (and Malnati's) course, and received a negative feedback on this proposal.

Therefore, we thank @umberto-fontanazza for his contribute, but we ask @QcFe to close this PR without merging.

AleTopp:
Furthermore, we would like to add that there is also a default shortcut by VSCode to format the current document which is `Alt+Shift+F`.

PR #972, opened 2025-09-13
Fix ShVol mount permissions for Workspace managers

AleTopp:
# Description

This PR makes the instctrl mount all Shared Volumes with RW access for workspace managers, since now there is no (obvious) way for a manager to add material to a RO drive.
It also includes a refactor that generalizes and moves the remaining NFS logic into the `forge` package.

Fixes issue raised by @air-31 (2025-09-10) and @frisso (2025-04-01)

~ Alessio Giliberti, Chiara Mercurio

# How Has This Been Tested?

Not tested yet

### DIFF
--- a/operators/pkg/forge/nfs.go
+++ b/operators/pkg/forge/nfs.go
@@ -17,6 +17,7 @@
 package forge
 
 import (
+	"context"
 	"fmt"
 	"maps"
 	"strings"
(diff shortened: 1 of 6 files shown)


AleTopp:
Test on staging completed. Everything is fine and working as expected.
All apparent problems were caused by a "corrupted" VM image (`annanarratone/sid2-prova`).
@QcFe you can merge.

AleTopp:
With this I'd like to point out to @air-31 that documentation about creating VMs should be updated consequently about the need of said package, to avoid similar problems in the future.

PR #975, opened 2025-09-30
fix default crowmail values

tollsimy:
# Description

Fix default CrownMails configs.


frisso:
For me, both options are valid.
I would also involve @MarcoRiki , as his new PR looks related to something that is being addressed in this PR as well (I can see some timeout changed in the code). Maybe, we can converge on just one PR that fixes the problems we encountered in the last period.
Thanks!

MarcoRiki:
I can work on this PR and delete the new one

air-31 (changes requested):
Apparently, when transferring some flags from `instance-operator` to `instance-operator-automation` deployment, the flag for shared volumes got deleted, so now ShVols have non-existent PVC storage class

QcFe (approved):

PR #988, opened 2025-10-28
ImageList handler

isadseta:
# Description
This PR rewrites the original Python script in Go, introducing a modular design and a thin main that only bootstraps configuration, wiring, and lifecycle. The binary periodically aggregates images from one or more requestors and writes/updates the ImageList CRD in Kubernetes.

# What changed
This pull request removes the standalone `crownlabs-image-list` Python service and fully integrates image list management into the main operator. Documentation has been updated to reflect the new architecture and usage.

# Rationale / Modularity

- main is only a bootstrap: parse flags → build requestors → construct 
- ImageListUpdater → run.
- Behavior is injected via interfaces (ImageListRequestor) and simple config (flags), which simplifies testing and future extensions (e.g., add a new requestor without touching main).
- Graceful shutdown implemented
- Requestor interface + registry pattern
- Added unit tests

# Results

As shown the list of images is stored successfully:
	 
image


QcFe:
IT ACTUALLY DID IT NOBODY BELIEVES

operators/cmd/operator/imagelist.go
```diff
@@ -0,0 +1,85 @@
...
+	flag.StringVar(&configFile, "config-file", "/etc/config/registries.yaml", "Path to the registries configuration file")
+	flag.IntVar(&updateInterval, "interval", 60, "Update interval in seconds")
+
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	ctrl.SetLogger(klog.NewKlogr())
+	log := ctrl.Log.WithName("imagelist")
+
+	k8sClient, err := utils.NewK8sClient()
```
QcFe: it's better to pass the mgr here and get logger & client from there

operators/build/crownlabs-image-list/Dockerfile
QcFe, line 1: this file can be removed

operators/cmd/operator/main.go
```diff
@@ -77,11 +79,19 @@ func main() {
...
 	flag.BoolVar(&enableWorkspace, "enable-workspace", true, "Enable the workspace controller.")
 	flag.BoolVar(&enableInstance, "enable-instance", true, "Enable the instance controller.")
 	flag.BoolVar(&enableSharedVolume, "enable-sharedvolume", true, "Enable the sharedvolume controller.")
 	flag.BoolVar(&enableKeycloak, "enable-keycloak", true, "Enable the Keycloak integration.")
+	flag.BoolVar(&enableImageList, "enable-image-list", true, "Enable the image list updater.")
+
+	// Image list configuration
+	var imageListConfigFile string
+	var imageListUpdateInterval int
+	flag.StringVar(&imageListConfigFile, "image-list-config-file", "/etc/config/registries.yaml", "Path to the image list registries configuration file")
```
QcFe: I'd keep flags population inside the other main like the other components do

operators/deploy/operator/templates/clusterrole.yaml
```diff
@@ -6,15 +6,15 @@ metadata:
...
+  resources: ["workspaces", "workspaces/status", "tenants", "tenants/status", "instances", "instances/status", "templates", "templates/status", "imagelists", "imagelists/status"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "deletecollection"]
 
 - apiGroups: ["crownlabs.polito.it"]
   resources: ["sharedvolumes", "sharedvolumes/status"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "deletecollection"]
   
 - apiGroups: [""]
-  resources: ["namespaces", "resourcequotas", "limitranges", "secrets"]
+  resources: ["namespaces", "resourcequotas", "limitranges", "secrets", "configmaps"]
```
QcFe: this shouldn't be needed

operators/deploy/operator/templates/configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: registry-configuration-image-list-update
```
QcFe:
```suggestion
  name: image-lists-sources
```

operators/deploy/operator/templates/deployment.yaml
```diff
@@ -96,10 +99,18 @@ spec:
           volumeMounts:
           - mountPath: {{ .Values.webhook.deployment.certsMount | default "/tmp/k8s-webhook-server/serving-certs/" }}
             name: webhook-certs
+          - mountPath: /etc/config
+            name: image-list-config
       volumes:
       - name: webhook-certs
         secret:
           secretName: {{ include "operator.webhookname" . }}
+      - name: image-list-config
```
QcFe, line 116: would mark this cm mounting as optional
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#optional-configmaps
(maybe also the webhooks)

operators/pkg/forge/utils.go
```diff
@@ -53,6 +54,14 @@ func ObjectMetaWithSuffix(instance *clv1alpha2.Instance, suffix string) metav1.O
 	}
 }
 
+// NamespacedNameImageList returns the namespace/name pair given an ImageList instance.
+func NamespacedNameImageList(instance *clv1alpha1.ImageList) types.NamespacedName {
+	return types.NamespacedName{
+		Name:      CanonicalName(instance.GetName()),
+		Namespace: instance.GetNamespace(),
```
QcFe: imagelists are cluster wide resources, don't think this helper is needed

operators/pkg/imagelist/agent.go
```diff
@@ -0,0 +1,231 @@
...
+type RegistryConfig struct {
+	Name          string `yaml:"name" json:"name"`
+	Type          string `yaml:"type" json:"type"`
+	URL           string `yaml:"url" json:"url"`
+	Advertised    string `yaml:"advertised" json:"advertised"`
+	Username      string `yaml:"username" json:"username"`
+	Password      string `yaml:"password" json:"password"`
+	ImageListName string `yaml:"imageListName" json:"imageListName"`
+	Project       string `yaml:"project" json:"project"`
+}
```
QcFe, line 40: it's also possible to use json only, helm is able to automatically translate yaml to json

MarcoRiki:
Update:
There are two imageLists on CrownNG: one for the standalone and the second for containerdisks images.
Argo has been updated, so now each environment (staging etc..) has its own imageLists: crownlabs-standalone-{environment}, crownlabs-containerdisks-{environment} ecc..
These names are actually set in the `image-lists-sources` CM which is deployed in the same ns of the environment

operators/pkg/imagelist/requestors.go
```diff
@@ -0,0 +1,609 @@
...
+// NewDockerImageListRequestor creates a new DockerImageListRequestor instance.
+func NewDockerImageListRequestor(log logr.Logger) *DockerImageListRequestor {
+	return &DockerImageListRequestor{
+		url:         "",
+		username:    "",
+		password:    "",
+		client:      &http.Client{},
+		initialized: false,
+		log:         log,
+	}
```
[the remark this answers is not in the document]
QcFe: @MarcoRiki if applyable this doesn't seem a particularly bad idea... maybe even just 10 secs or something?

operators/pkg/imagelist/imgUpdater.go
```diff
@@ -0,0 +1,116 @@
...
+						versions = append(versions, s)
+					}
+				}
+			}
+		}
+
+		// Ensure at least one version exists
+		if len(versions) == 0 {
+			versions = []string{"latest"}
+		}
```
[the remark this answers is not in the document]
QcFe: @MarcoRiki this isn't a particularly wrong point :(
since we're merging this in the main operator, the risk is to add untested logic that could possibly break (mostly, a possible tomorrow) a rather large component...

operators/pkg/imagelist/imgUpdater.go
```diff
@@ -0,0 +1,116 @@
...
+
+		out = append(out, clv1alpha1.ImageListItem{
+			Name:     name,
+			Versions: versions,
+		})
```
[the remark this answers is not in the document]
MarcoRiki: @QcFe I don't know what to do here. The old script was not appending the images if they had "latest" as tag. It was appending only versioned images but the "latest" was excluded. Here the behaviour is quite different: if the image has only the "latest" version, then it is kept; while if versioned tags are available, the "latest" one is again excluded. Is it ok?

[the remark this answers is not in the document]
QcFe: I'd personally say let's exclude latest in any case. If there's latest only, the image is not to be considered.

operators/pkg/imagelist/imgUpdater.go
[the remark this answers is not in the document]
MarcoRiki: Ok I can add some tests, at least for the most critical functions such as the processing/updating of the imageList and the connection to the repo. What do you think?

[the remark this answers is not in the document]
QcFe: yeah that's reasonable. I think copilot in this context can be worth a try in any case

QcFe (approved):

PR #989, opened 2025-10-30
bugfix: tenant namespace defined incorrectly

SebastianMontoyaZ:
# Description

Bugfix for error introduced in [merge](https://github.com/netgroup-polito/CrownLabs/pull/968) that prevented users with specific tenant names from creating instances

Fixes # (issue)

# How Has This Been Tested?

local testing with differently named tenants


QcFe:
Ciao @SebastianMontoyaZ, grazie. I'd ask @giando332994 if you can report here the other issue you notice.

giando332994:
In the Active View the template name is not displayed and the error should be here
image

giando332994:
also I have an error on the sharedvolumes when i open any workspace in the gui
image
in the active view in the managed section I can't see correctly the row: Status ID User Instance Name  Utils Age Actions on top of every instance
image

Issue #992, opened 2025-11-04
[Feature] Removing a user from a workspace


abellonioandrea:
Now it's only possibile to add user in workspace and edit the role, but the removal is not implemented yet.

PR #993, opened 2025-11-04
feat(user management): removing a user from a workspace

abellonioandrea:
# Description
Adding the functionality of removing users from workspace. Before the button was present in the users management section, but not implemented.

Fixes issue #992

# How Has This Been Tested?
Tested with the gui: adding and removing users is working.


frisso:
@abellonioandrea This PR is ready for review, isn't it?

PR #1008, opened 2025-11-27
adding inactivityTimeout and deleteAfter fields in the frontend template creation

MarcoRiki:
This PR is an extension of https://github.com/netgroup-polito/CrownLabs/pull/934.
That PR added two new fields in the Template CRD:

inactivityTimeout -> Instances based on this template are stopped / deleted if they're not accessed within this time.
deleteAfter -> Time, since the creation, after which instances based on this template are automatically deleted. Users will be preemptively notified via email to take action.
This PR extends the frontend to be able to set these two fields.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.2.0",
       "dependencies": {
         "@ant-design/cssinjs": "^1.23.0",
-        "@ant-design/icons": "^6.1.0",
+        "@ant-design/icons": "^5.6.1",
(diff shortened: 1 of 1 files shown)


frontend/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "dependencies": {
     "@ant-design/cssinjs": "^1.23.0",
-    "@ant-design/icons": "^5.6.1",
+    "@ant-design/icons": "^6.1.0",
```
frisso:
```suggestion
    "@ant-design/icons": "^5.6.1",
```

Issue #1015, opened 2025-12-01
[Bug] Resource consumption panel reports wrong numbers in case of stopped instances


frisso:
The resource consumption is reported wrongly in case of stopped instances. For instance, stopped instances do not consume resources, so their allocation should not be counted in the total amount of resources consumed per user.
The following screenshot shows the misbehavior: all instances are stopped, while the resource dashboard at the bottom reports 3 instances running.

<img width="1397" height="854" alt="Image" src="https://github.com/user-attachments/assets/a1e71646-ab2b-43be-9329-a62ddd29c976" />

**Note from Federico**: way we currently have 3 different places that all retrieve instances and basically share no code between each other. Likely, a refactor would be less painful than reimplementing possible filtering logic or similar

AleTopp:
I noticed the same thing. Maybe that information could also be obtained fetching directly the ResourceQuota (that is what that report wanted to convey):
<img width="2422" height="216" alt="Image" src="https://github.com/user-attachments/assets/97d457db-5cf7-439b-8756-9851f6382e7a" />

About the Note, there is already an issue open on that: #1009

Issue #1009, opened 2025-11-27
Frontend: owned instances refactor


QcFe:
Currently owned instances are fetched and handled in too many places. It's the case to refactor in order to make a single, shared context that can be then used to render various elements in the ui.

Alternatively, we can see if there's any way to use something like tanstack query to make the whole data/state management more meaningful.

PR #1021, opened 2025-12-09
Owned instances refactoring

abellonioandrea:
# Description
Owned instances refactoring: now owned instances are managed in only one react context shared across the frontend.
fix(quotas): now only the running instances are used for displaying consumed quotas (paused instances doesn't count)

# Fixes issues

- #1015
- #1009 

# How Has This Been Tested?
Tested with GUI locally.


AleTopp:
I'm sorry I didn't attend that discussion, that's fine then. I was erroneously thinking that stopped VMs don't come for free but all that is kept offline is on the PV that would be counted against the disk quota :)
Then, I agree it should be done in a separate PR since this was for refactoring.

riccardotornesello:
> Nice observation, @AleTopp However, to solve this issue, shouldn't be better to add also disk quota? It could be a nice task to @riccardotornesello.

That would be a nice update. I'd prefer to give priority to my update about granular workspace quotas and then proceed with the improvements in subsequent PRs

Issue #962, opened 2025-08-23
Inconsistency in logic of handling resource quotas for tenants and workspaces


air-31:
Currently, the resource quota of each tenant is the result of aggregation of resource quotas of all workspaces a tenant is assigned to. However, a user is free to consume all the resources assigned to them within a single workspace.

[Proposal] To substantiate the idea behind workspace resource quotas, it is suggested to make them take precedence over tenant resource quotas, so that no (general) user can overcome the limitations set in workspace specifications while working inside that workspace. Meaning, if the workspace limit for the number of instances is 1, it should not be possible to launch a second instance from any template of that workspace.

[Side note] There are resource caps per tenant for each metric of resource quotas. To update or overcome those caps is not so straightforward as of now. To make the handling of the resource caps more manageable is to be considered.

AleTopp:
Solved by #1022

PR #1022, opened 2025-12-12
[Feature] Granular workspace quotas

riccardotornesello:
# Description

The main objective of this change is to ensure that quotas are divided for each workspace.
Currently, users can use resources up to the total amount of resources available in all the workspaces they belong to, ignoring the limits of individual workspaces.

This change aims to implement a control on the creation of a new instance to ensure that it does not exceed the limits of the workspace itself, considering the other environments already running.

The change will also affect the management of tenants' personal workspaces: it will be necessary to specify which resources the user can use in that specific workspace.

Fixes Issue #962 

# Important notes and breaking changes

1. The updates to the CRDs require kubernetes >= 1.29.0 due to the CEL syntax
2. All the personal workspaces will be disabled since the old field in the Tenant used to enable it will be replaced by a field that specifies the workspace's quota

# ~~Dependencies~~

~~The frontend updates are ready but saved in a "temporary" branch, since they depend on PR #1021. Once that PR will be merged, this branch will be rebased and the UI updates will be added here.~~

# Changes

## Webhook

Instance validation is performed by a new webhook added to the “operator” controller. NOTE: the maintainers explicitly asked to implement it in the operator controller since the instance-operator will be merged there.
This webhook receives requests to create new instances and checks that the tenant has enough resources available in that specific workspace for the new instance.
Furthermore, since the resources of paused instances are not counted, it also intercepts update requests, validating resources only if the instance changes from a spec with “running: false” to “running: true.”

## Personal workspace

The management of personal workspaces has been completely redesigned.
The “createPersonalWorkspace” field has been removed, while the ‘quota’ field has become “personalWorkspaceQuota” and is mandatory for enabling a tenant's personal workspace, so that you always have control over the resources assigned to it.

In addition, to make the code more uniform, the TenantResourceQuota struct has been replaced by the WorkspaceResourceQuota struct, which already exists and is used for normal workspaces.

## ResourceQuota

Kubernetes ResourceQuota resources remain in use as a final security barrier, but their calculation has changed.
Since the tenant quota field is no longer an override but rather the amount of resources assigned to its personal workspace, these are added to those of the individual workspaces to which it belongs and do not replace them.

## Show quotas in frontend

The frontend has been modified to show in the dashboard how many resources have been used in individual workspaces rather than globally (it is responsive ;)).

In addition, the code for calculating remaining quotas has been cleaned up and updated. To do this, QuotaContext and useQuotaCalculation have been removed, moving the calculation to the OwnedInstancesContext context.

image

## Manage tenant's personal workspace

System administrators can access an additional dashboard that allows them to view tenant details and activate, deactivate, or modify their personal workspace.

NOTE: To view this panel, you must be part of the “admin” group in Keycloak.

image

## Removed users management from personal workspace

The button on the left (Manage Users) in the Personal Workspace's card has been removed.

image

## Other minor changes

- Increased operator memory to 1GB

# Follow-ups

## Disk quota

Disk quota management is currently set up in the frontend, but TODOs have been left where changes are needed to implement the functionality.

### DIFF
--- a/frontend/src/components/workspaces/WorkspaceContainer/WorkspaceContainer.tsx
+++ b/frontend/src/components/workspaces/WorkspaceContainer/WorkspaceContainer.tsx
@@ -129,24 +129,26 @@ const WorkspaceContainer: FC<IWorkspaceContainerProps> = ({ ...props }) => {
             <div className="h-full flex justify-center items-center flex-col">
               <div className="w-full flex-grow flex justify-center items-center">
                 {workspace.role === WorkspaceRole.manager && (
-                  <div className="h-full flex-none flex justify-center items-center pl-10">
-                    <Tooltip title="Manage users">
-                      <Button
-                        type="primary"
-                        shape="circle"
-                        size="large"
-                        icon={<UserSwitchOutlined />}
-                        onClick={() => setShowUserListModal(true)}
-                      >
-                        {workspace.waitingTenants && (
-                          <Badge
-                            count={workspace.waitingTenants}
-                            color="yellow"
-                            className="absolute -top-2.5 -right-2.5"
-                          />
-                        )}
-                      </Button>
-                    </Tooltip>
+                  <div className="h-full w-18 md:w-24 flex-none flex justify-center items-center">
+                    {!isPersonal && (
+                      <Tooltip title="Manage users">
+                        <Button
+                          type="primary"
+                          shape="circle"
+                          size="large"
+                          icon={<UserSwitchOutlined />}
+                          onClick={() => setShowUserListModal(true)}
+                        >
+                          {workspace.waitingTenants && (
+                            <Badge
+                              count={workspace.waitingTenants}
+                              color="yellow"
+                              className="absolute -top-2.5 -right-2.5"
+                            />
+                          )}
+                        </Button>
+                      </Tooltip>
+                    )}
                   </div>
                 )}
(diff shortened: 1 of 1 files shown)


riccardotornesello:
Thanks @abellonioandrea.
I just pushed a commit to fix this issue.
In the `utilsLogic.tsx` file, there was a function named getManagerInstances which was similar but not equal to `makeGuiInstance`. I decided to remove `getManagerInstances` and replace it with `makeGuiInstance`.

deploy/crownlabs/Chart.lock
```diff
@@ -5,18 +5,18 @@ dependencies:
 - name: qlkube
   repository: file://../../qlkube/deploy/qlkube
   version: 0.1.0
+- name: operator
```
frisso: @QcFe I know I'm rather late now, but the name "operator" looks confusing to me. Can we define a better name? Even "crownlabs-operator" looks more appropriate to me.
If you agree, we can think about merging this PR "as is", then create another that just makes the rename.

QcFe:
> we thought it was a business decision

More than a business decision I think it was also because of implementation simplicity, I think the RBAC part hasn't been ported to the personal workspace. Indeed I'm not extremely sure how useful it can be with the myDrive that can be already attached to every instance, risk could be for those users to saturate the storage more than what they already could 😬

PR #1037, opened 2026-01-27
refactor lastLogin field in Tenant resourse

air-31:
# Description

This PR aims to make the `spec.lastLogin` field of the Tenant custom resource to follow the `omitempty` behaviour, because even when omitting it during the Tenant creation, the request sent to the API server still contains the field as equal to null

### DIFF
--- a/operators/pkg/controller/tenant/namespace.go
+++ b/operators/pkg/controller/tenant/namespace.go
@@ -163,14 +163,13 @@ func (r *Reconciler) checkNamespaceKeepAlive(ctx context.Context, log logr.Logge
 	// We check to see if last login was more than r.TenantNSKeepAlive in the past:
 	// if so, temporarily delete the namespace. The lastLogin field is omitted when a user is first created
 
-	var sPassed time.Duration
+	sPassed := r.TenantNSKeepAlive + (1 * time.Second)
 
 	if tn.Spec.LastLogin != nil {
 		// Calculate time elapsed since lastLogin (now minus lastLogin in seconds)
 		sPassed = time.Since(tn.Spec.LastLogin.Time)
 		log.Info("Last login checked", "tenant", tn.Name, "elapsed", sPassed)
 	} else {
-		sPassed = r.TenantNSKeepAlive + (1 * time.Second)
 		log.Info("No last login found: assuming tenant inactive", "tenant", tn.Name)
 	}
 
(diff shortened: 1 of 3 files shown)


air-31:
@QcFe I had to make small changes in `validator.go` and `namespace.go` which I'm not extremely sure about, so I'd appreciate you taking a look whenever you have time before finalizing (not urgent)

operators/pkg/controller/tenant/namespace.go
```diff
@@ -161,12 +161,18 @@ func (r *Reconciler) enforcePersonalNamespaceAbsence(
 // checkNamespaceKeepAlive checks to see if the namespace should be deleted.
 func (r *Reconciler) checkNamespaceKeepAlive(ctx context.Context, log logr.Logger, tn *v1alpha2.Tenant) (keepNsOpen bool, err error) {
 	// We check to see if last login was more than r.TenantNSKeepAlive in the past:
-	// if so, temporarily delete the namespace. We assume that a lastLogin of 0 occurs when a user is first created
+	// if so, temporarily delete the namespace. The lastLogin field is omitted when a user is first created
 
-	// Calculate time elapsed since lastLogin (now minus lastLogin in seconds)
-	sPassed := time.Since(tn.Spec.LastLogin.Time)
+	var sPassed time.Duration
```
QcFe: probably it's better to pre-set it to a big time, like, 100 years or something

operators/pkg/controller/tenant/webhook/validator.go
```diff
@@ -237,12 +236,26 @@ func (tv *TenantValidator) HandleSelfEdit(
 	newTenant.Spec.PublicKeys = nil
 	oldTenant.Spec.PublicKeys = nil
 
-	lastLoginDelta := time.Until(newTenant.Spec.LastLogin.Time).Abs()
-	if newTenant.Spec.LastLogin != oldTenant.Spec.LastLogin && lastLoginDelta > LastLoginToleration {
-		return nil, errors.NewForbidden(schema.GroupResource{}, newTenant.Name, fmt.Errorf("you are not allowed to change the LastLogin field in the owned tenant, or the change is not valid: %s", lastLoginDelta))
+	// manage last login
```
QcFe: I'm not so sure it's so needed to check if it has been changed, I'd just verify that the value in the new one is compliant: if the old one was wrong (and it's unchanged), I'd complain still

Issue #1027, opened 2025-12-18
[Feature] Use Template description in the frontend


AleTopp:
**Is your feature request related to a problem? Please describe.**
The Template CRD has a [`Description` field](https://github.com/netgroup-polito/CrownLabs/blob/0d0d3a33e72aea3867cbde76fb4087329a6af860/operators/api/v1alpha2/template_types.go#L58C1-L58C41), that is set for most of the templates (since they were manually applied to k8s), but unused.

**Describe the solution you'd like**
 I would like to see the Template description somewhere in the frontend, and at same time it should be possible for teachers to set it (at creation) and update it.

**Additional context**
During these semesters' labs I heard many times "which VM should I start?", so maybe including the description somewhere in the frontend could be an idea to solve this issue.

Issue #987, opened 2025-10-28
[Feature] Implement public exposure flag on template editor (frontend)


QcFe:
**Describe the solution you'd like**
A checkbox in the template editor page.

PR #1043, opened 2026-01-30
Extend template editor to support new features (frontend) 

MarcoRiki:
PR based on the issue #987

The CrownLabs backend has been recently enriched with new features, which are not yet supported by the frontend. Hence, these features can be enabled only by manually editing the template CRD.
In particular, the following features are missing:
- [x] capability to **publicly expose VMs/containers on a public IP address**, in order to allow users to use CrownLabs as "server" cloud facility. This feature is documented here: https://github.com/netgroup-polito/CrownLabs/tree/master/operators#instance-services-public-exposure
- [x] capability to **set the node on which an instance should be running**. This feature is documented here: https://github.com/netgroup-polito/CrownLabs/blob/master/operators/docs/node-selection.md
- [x] capability to set the **CPU requests** (or reserved %) per each template: the current setting of the CPU limit can bring to very high inefficiencies, hence the CPU request looks like a more balanced way to set a reasonable amount of CPU power to each instance, avoiding excessive resource waste.
- [x] possibility to set the [**_Description_ field**](https://github.com/netgroup-polito/CrownLabs/blob/0d0d3a33e72aea3867cbde76fb4087329a6af860/operators/api/v1alpha2/template_types.go#L58C1-L58C41), which can be useful to describe better the template we're creating (#1027).

This feature request proposes to extend the template editing window with the above new features.

In order to avoid increasing clutter onto the template editor:
- [x] Refactor the editor into accordion (antd: collapse component) sections:
  - Cleanup (including a pill saying if it's enabled or not)
  - Features (including pills for saying features status)
  - Environments (including a pill for each environment)
  *pills should be visible onto the collapse title in order to provide settings at a glance
- [x] Set frontend validation for environment length
- [x] remove resources sliders in favor to input boxex

### DIFF
--- a/.github/actions/retrieve-build-matrix/package-lock.json
+++ b/.github/actions/retrieve-build-matrix/package-lock.json
@@ -2334,9 +2334,9 @@
       }
     },
     "node_modules/minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
+      "version": "3.1.5",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.5.tgz",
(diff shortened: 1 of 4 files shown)


abellonioandrea:
Hello everyone,
I was testing this new feature. I created a template, after I edited that (to be specific the automatic clean up) and when confirming, this error showed up. Fell free to ask for any additional info.
image

Edit: I think that is a spread error over this modal, I found also other field giving error when edited (like template name, image name/path)

PR #1066, opened 2026-03-13
fix minor bugs for memory visualization in frontend

MarcoRiki:
# Description

Memory and Disk visualization is not consistent if the unit of measurement is set to G instead of M, leading to errors in frontend.


air-31:
@MarcoRiki thanks again!
I don't think it's necessary to fix that here, since this PR was intended to be a quick fix to make the editor usable for end users.
That said, let's wait for Federico's feedback

Issue #1059, opened 2026-02-26
[Feature] LIVE MIGRATION


QcFe:
**Is your feature request related to a problem? Please describe.**
VMs can't be live-migrated

**Describe the solution you'd like**
Switch PVCs to Block Mode with RWX (IT CAN BE ACTUALLY DONE)

**Describe alternatives you've considered**
Doing CephFS with RWX but everyone says it has bad performance.

**ALSO**
VMs should also work faster with an actual block volume instead of a virtual disk file onto a fs onto a block volume.

PR #1094, opened 2026-04-16
Implementation of LocalVM with PVC smart cloning

Deblazz:
# Description

_This PR is part of project 5 from Cloud Programming Course of ay 2025-26_

VolumeMode of CloudVMs virtual disks migrated from Filesystem to Block type, to have less layers of abstraction between the running VM and ceph.
New EnvironmentType created for LocalVM: now it's possibile to clone a PVC from another namespace, using smartcloning implemented in CDI. Also, the frontend has been modified to implement some basic features for using this new mode.
ImageName is specified in the template as: _namespace/pvcname_

This PR contribute to achive a solution for the issue #1059 .

[https://github.com/kubevirt/containerized-data-importer/blob/main/doc/clone-datavolume.md](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/clone-datavolume.md)
[https://github.com/kubevirt/containerized-data-importer/blob/main/doc/smart-clone.md](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/smart-clone.md)

# How Has This Been Tested?

This configuration has been functionally tested by running a VM successfully.


AleTopp:
@Deblazz @abellonioandrea Sorry, I think I lost something, didn't you say on Tuesday that changing to Block was increasing by 100% the startup time?

abellonioandrea:
We are doing some tests, since with a loop disk (so that is not reflecting correctly the reality) seems that the time needed is higher. In any case, another point of our project is to allow Kubevirt VMs to be live migrated, and it is only possibile with the volume mode set to block and not with filesystem.

operators/pkg/forge/virtualmachines.go
```diff
@@ -247,7 +247,9 @@ func DataVolumeTemplate(name string, environment *clv1alpha2.Environment) virtv1
 		Spec: cdiv1beta1.DataVolumeSpec{
 			Source: DataVolumeSourceForge(environment),
 			PVC: &corev1.PersistentVolumeClaimSpec{
-				AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+				VolumeMode:       ptr.To(corev1.PersistentVolumeBlock),
+				AccessModes:      []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
```
QcFe: I think this needs to be RWX or we won't be able to do live migration :) 
Ceph underneath supports RWX for block volumes

abellonioandrea: Yes it is necessary, we updated the code accordingly
Thanks for the feedback

QcFe:
I'd say this can be almost merged, would just ask to write a brief test for your new method.
When you're done don't forget to set this PR to ready :)

abellonioandrea:
I have just added an annotation, and now live migration it's working.
We will wrote some tests!

operators/pkg/forge/virtualmachines.go
```diff
@@ -54,8 +58,13 @@ var (
 func VirtualMachineSpec(instance *clv1alpha2.Instance, template *clv1alpha2.Template, environment *clv1alpha2.Environment) virtv1.VirtualMachineSpec {
 	return virtv1.VirtualMachineSpec{
 		Template: &virtv1.VirtualMachineInstanceTemplateSpec{
-			ObjectMeta: metav1.ObjectMeta{Labels: EnvironmentSelectorLabels(instance, environment)},
-			Spec:       VirtualMachineInstanceSpec(instance, template, environment),
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: EnvironmentSelectorLabels(instance, environment),
+				Annotations: map[string]string{
+					PodBridgeNetworkLiveMigrationAnnotation: podBridgeNetworkLiveMigrationValue,
```
QcFe: we're theoretically not using bridge mode, is this really needed?

abellonioandrea: Without this, the live migration is not working
To be precise in the virtual machine I am getting this error:
`cannot migrate VMI which does not use masquerade, bridge with kubevirt.io/allow-pod-bridge-network-live-migration VM annotation or a migratable plugin to connect to the pod network`

QcFe: ok but we're supposed to be using masquerade actually...

abellonioandrea: okay I didn't know, so it is only a problem related to the fact we are running vms manually in KubeVirt without the crownlabs infrastructure

abellonioandrea: I will revert the changes

QcFe: can you please check just in case? I added self-edit on your tenant ns.

abellonioandrea: I confirm that on a virtual machine in my tenant namespace I see the same error as before for the live migration enable

abellonioandrea: I think that actually VMS are not in Masquerade but are in Bridge mode.
https://github.com/netgroup-polito/CrownLabs/blob/620c972bea0d97281c69e0f90627c615969cbc64/operators/pkg/forge/virtualmachines.go#L88
DefaultBridgeNetworkInterface must be changed to DefaultMasqueradeNetworkInterface.

QcFe: this explains several things. I'd say it's better to discuss this tonight at the weekly in order to understand together if there could be further implications by switching this...

AleTopp:
Someone is ded
immagine

AleTopp:
I created a new PR with the commits about PMP that were in this branch. You can now rebase (deleting the useless commits) and squash!

Deblazz:
I think we should be good to go. I'm available for further necessities!

QcFe:
> I think we should be good to go. I'm available for further necessities!

Ok thanks! I'd ask you to rebase with main then set the pr to ready when you want me to have a look ;)

operators/pkg/forge/virtualmachines.go
```diff
@@ -256,15 +259,42 @@ func VirtualMachineReadinessProbe(environment *clv1alpha2.Environment) *virtv1.P
...
+		}
+		// TODO: Manage error! The inserted Image is not valid as it contains more than one "/"
+		// Currently defaults to the debian pvc
+		return &cdiv1beta1.DataVolumeSource{
+			PVC: &cdiv1beta1.DataVolumeSourcePVC{
+				Namespace: "cldprog-5-block-vms-tests",
+				Name:      "debian-nginx-raw-block",
+			},
+		}
+	}
```
QcFe: I'd say it's better to solve this todo before merging :)

operators/pkg/forge/virtualmachines.go
```diff
@@ -273,14 +303,25 @@ func DataVolumeSourceForge(environment *clv1alpha2.Environment) *cdiv1beta1.Data
 	}
 }
 
-// DataVolumeTemplate forges the DataVolume template associated with a given environment.
-func DataVolumeTemplate(name string, environment *clv1alpha2.Environment) virtv1.DataVolumeTemplateSpec {
-	return virtv1.DataVolumeTemplateSpec{
-		ObjectMeta: metav1.ObjectMeta{Name: name},
+// DataVolume forging, it needs to be created before the VM, forged with the same name as the VM.
+func DataVolume(name, namespace string, environment *clv1alpha2.Environment) *cdiv1beta1.DataVolume {
```
QcFe: iirc we said to switch this one to DataVolumeSpec to keep it streamlined with the other forge functions

QcFe:
Considering the main instance ctrl logic has been changed about vms, I'd prefer if you could possibly add some tests about it in the corresponding file

QcFe (approved):

PR #1099, opened 2026-04-21
PMP Mirror Enforcement

chia-me:
# Description

This PR is the continuation of #1067, and aims to enforce mirrors for MyDrives and SharedVolumes, and their mounting on the instances, thus removing NFS.

This is part of the Cloud Programming project no.3 of a.y. 2025/26, done by @chia-me and @AleTopp

# How Has This Been Tested?

Each function added or edited was unit-tested, and integration-tested as the methods around them.


operators/pkg/controller/sharedvolume/reconciler.go
```diff
@@ -147,7 +147,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 	shvolume.Status.Phase = clv1alpha2.SharedVolumePhaseError
 
 	// Create or Update the PVC, reconciling it with the SharedVolume spec.
-	pvc := v1.PersistentVolumeClaim{ObjectMeta: metav1.ObjectMeta{Name: "shvol-" + shvolume.GetName(), Namespace: shvolume.GetNamespace()}}
+	pvc := v1.PersistentVolumeClaim{ObjectMeta: metav1.ObjectMeta{
+		Name:      forge.ShVolPVCName(shvolume.GetName()),
+		Namespace: shvolume.GetNamespace(),
+	}}
```
QcFe, line 153: there should be a forge helper to create namespaced names, should possibly be used in any other case too...

operators/pkg/forge/cloudinit.go
```diff
@@ -93,3 +94,21 @@ func CloudInitUserData(publicKeys []string, mountInfos []NFSVolumeMountInfo) ([]
 	output = append([]byte("#cloud-config\n"), output...)
 	return output, nil
 }
+
+// virtiofsVolumeMount forges the mount string array for a generic Virtiofs volume.
+func virtiofsVolumeMount(mount *corev1.VolumeMount) []string {
+	rwPermission := "rw"
```
QcFe: would call this permission only maybe

operators/pkg/forge/virtualmachines.go
```diff
@@ -78,28 +78,54 @@ func VirtualMachineInstanceSpec(instance *clv1alpha2.Instance, template *clv1alp
...
+func AttachableVolumes(mountInfos []corev1.VolumeMount) []virtv1.Volume {
+	volumes := []virtv1.Volume{}
+	for _, mount := range mountInfos {
+		volumes = append(volumes, VolumePVC(&mount))
+	}
+	return volumes
+}
+
+// VolumePVC forges the specification of an external volume (MyDrive, SharedVolume) to be mounted through PVC.
+func VolumePVC(mount *corev1.VolumeMount) virtv1.Volume {
```
QcFe:
```suggestion
func VirtPVCVolume(mount *corev1.VolumeMount) virtv1.Volume {
```
would specify is vms stuff just in case since we're in a general forge pkg here

operators/pkg/forge/cloudinit.go
AleTopp: wdym?

operators/pkg/instctrl/storage.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+			mirrPvc.SetLabels(forge.UpdateShVolMirrorPVCLabels(mirrPvc.Labels))
+
+			return controllerutil.SetControllerReference(instance, &mirrPvc, r.Scheme)
+		})
+		if err != nil {
+			log.Error(err, "failed to enforce mirror sharedvolume", "SharedVolume", shvolPvc, "Environment", environment)
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
AleTopp: Instances will not be reconciled to the Template spec after creation, so even if some ShVol was unmounted from the template, it would still be mounted on the instance.
Furthermore, even if there was some orphaned mirror, they would be deleted at the instance deletion since they are owned by it.

operators/cmd/instance-operator/main.go
```diff
@@ -98,6 +99,8 @@ func main() {
 	flag.StringVar(&publicExposureCommonLabelsRaw, "public-exposure-common-labels", "", "Comma-separated list of common labels in format key1=val1,key2=val2")
 	flag.StringVar(&publicExposureOpts.LoadBalancerIPsKey, "public-exposure-loadbalancer-ips-key", "metallb.universe.tf/loadBalancerIPs", "Annotation key for specifying LoadBalancer IPs")
 
+	flag.StringVar(&mirrorStorageClass, "mirror-storage-class", "nfs-mirror", "The StorageClass to be used for all PVCs which are going to be mirrors")
```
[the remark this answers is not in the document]
AleTopp: Yes already done

operators/pkg/controller/sharedvolume/reconciler.go
chia-me: Actually, almost everywhere there is an inline creation of ObjectMeta, only for some specific cases (like for some Instance/Environment enforcement) there are forge functions for NamespacedNames

chia-me: Furthermore, this would be the only place that would use it, and you cannot assign types.NamespacedName to metav1.ObjectMeta

operators/pkg/forge/cloudinit.go
QcFe: the rwPermission var :) when you set it to ro it's not rw anymore

operators/pkg/controller/sharedvolume/reconciler.go
QcFe: Hmm I think I this is because the forge.ObjectMeta method was made to receive an instance, however your use-case is mostly identical... If you switch that func to receive a `metav1.Object` instead of an instance, then you'd be able to use it too... (would do this for every function that uses Instance, where applicable, in forge/utils.go...)

AleTopp: We agree but we think such a cleanup would be better in a separate PR (maybe the one we are already preparing, since it refactors other things like NFS provisioning)

QcFe (approved):
if you can have one last quick check at the staging, then we're good to go

PR #1111, opened 2026-05-11
Fix indefinitely Deleting SharedVolumes

chia-me:
# Description
This PR aims to fix the behaviour of `SharedVolumes` that stay in the Deleting phase as long as they don't get reconciled, even if the Template that was blocking them has been deleted or has unmounted it.
This fix consists in adding a Watch on `Template`s to the `shvol-ctrl`, that whenever a `Template` is created/updated/deleted, the controller will reenqueue a reconcile request for all Deleting `SharedVolumes`, that will be deleted if it's possible now.

Co-authored by @AleTopp 

# How Has This Been Tested?
Only tested on the staging environment.


operators/pkg/controller/sharedvolume/reconciler.go
```diff
@@ -276,3 +295,34 @@ func (r *Reconciler) handleDeletion(ctx context.Context, log logr.Logger, shvol
...
+// getDeletingSharedVolumes returns reconciliation requests for SharedVolumes in Deleting phase so they can be reenqueued,
+// since the SharedVolumes cannot be deleted until all Templates have removed their mounts.
+func (r *Reconciler) getDeletingSharedVolumes(ctx context.Context, obj client.Object) []ctrl.Request {
+	var enqueues []ctrl.Request
+	var shvols clv1alpha2.SharedVolumeList
+
+	log := ctrl.LoggerFrom(ctx, "woken-by-template", types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	})
```
QcFe, line 308: Wasn't there was a helper to make a NamespacedNames?

QcFe (approved):

operators/pkg/controller/sharedvolume/reconciler.go
AleTopp: We already talked about it, it will be refactored in another PR by adding a generic function and using it here and in every other place it's done this way!

QcFe: :')

PR #1125, opened 2026-05-17
Cleanup P7: Remove unnecessary Template Scope logic

alexannekk:
This pull request removes all code paths that rely on the Scope field in CrownLabs templates.

While the field is present in the CRD, it is not used. As a result, the conditional logic tied to ScopeStandard, ScopeExercise, and ScopeExam has become unnecessary.

The cleanup primarily affects the forge package and the instance‑controller, where all scope‑dependent branches have been removed. Additionally, the Scope attribute has been fully dropped from the TemplateSpec in the v1alpha2 API version.


operators/pkg/forge/virtualmachines.go
```diff
@@ -78,30 +81,19 @@ func VirtualMachineInstanceSpec(instance *clv1alpha2.Instance, template *clv1alp
...
 // object representing the definition of the VM corresponding to a given CrownLabs Environment.
-func VirtualMachineDomain(environment *clv1alpha2.Environment, template *clv1alpha2.Template, mountInfos []corev1.VolumeMount) virtv1.DomainSpec {
+func VirtualMachineDomain(environment *clv1alpha2.Environment, mountInfos []corev1.VolumeMount) virtv1.DomainSpec {
 	return virtv1.DomainSpec{
 		CPU:       &virtv1.CPU{Cores: environment.Resources.CPU},
 		Memory:    &virtv1.Memory{Guest: &environment.Resources.Memory},
 		Resources: VirtualMachineResources(environment),
 		Devices: virtv1.Devices{
-			Disks:       VolumeDiskTargets(environment, template),
+			Disks:       []virtv1.Disk{VolumeDiskTarget(volumeRootName)},
```
QcFe: I'm afraid a bit too many things have been removed here...

operators/pkg/instctrl/exposition.go
```diff
@@ -99,10 +99,6 @@ func (r *InstanceReconciler) enforceInstanceExpositionPresence(ctx context.Conte
 
 		ingressGUI.SetAnnotations(forge.IngressGUIAnnotations(environment, ingressGUI.GetAnnotations()))
 
-		if template.Spec.Scope == clv1alpha2.ScopeStandard {
```
QcFe, line 102: This was actually needed. Since the scope has been removed, authentication has to be still disabled when needed. We'd better discuss if to do this at operator deployment level (ie, through a flag, enough for our needs and safer) or per template (however it might be abused by workspace manager).

operators/pkg/forge/containers.go
```diff
@@ -153,14 +152,13 @@ func DeploymentSpec(instance *clv1alpha2.Instance, template *clv1alpha2.Template
 // PodSpec forges the pod specification for X-VNC based container instance.
 func PodSpec(instance *clv1alpha2.Instance, template *clv1alpha2.Template, environment *clv1alpha2.Environment, mountInfos []corev1.VolumeMount, opts *ContainerEnvOpts) corev1.PodSpec {
 	return corev1.PodSpec{
-		Containers:                    ContainersSpec(instance, environment, mountInfos, opts),
+		Containers:                    []corev1.Container{StandaloneContainer(instance, environment, PersistentMountPath(environment), mountInfos)},
```
QcFe, line 155: If ContainerSpec isn't used anymore it's better to remove it (couldn't find the removal)

operators/pkg/instctrl/virtualmachines.go
```diff
@@ -34,15 +34,6 @@ import (
 func (r *InstanceReconciler) EnforceVMEnvironment(ctx context.Context) error {
 	log := ctrl.LoggerFrom(ctx)
 	environment := clctx.EnvironmentFrom(ctx)
-	template := clctx.TemplateFrom(ctx)
-
-	// Enforce the cloud-init secret when environment is not restricted
-	if template.Spec.Scope == clv1alpha2.ScopeStandard {
-		if err := r.EnforceCloudInitSecret(ctx); err != nil {
```
QcFe, line 41: Also here, cloudinit needs to be always set

QcFe (commented):
This review is partial, please have a look also at the rest of the diffs, since many times I reckon a bit too much logic has been removed

QcFe (approved):

PR #1133, opened 2026-05-30
chore: refactor NamespacedName utility methods

AleTopp:
# Description

This PR aims to refactor the abundant creation of `types.NamespacedName` objects used to Get kubernetes objects, by generalizing already present utility functions.
 
# How Has This Been Tested?

Edited functions had their tests edited accordingly.

### DIFF
--- a/operators/pkg/forge/utils.go
+++ b/operators/pkg/forge/utils.go
@@ -53,14 +53,6 @@ func ObjectMetaWithSuffix(instance *clv1alpha2.Instance, suffix string) metav1.O
 	}
 }
 
-// NamespacedName returns the namespace/name pair given an instance object.
-func NamespacedName(instance *clv1alpha2.Instance) types.NamespacedName {
-	return types.NamespacedName{
-		Name:      CanonicalName(instance.GetName()),
-		Namespace: instance.GetNamespace(),
-	}
-}
-
 // NamespacedNameWithSuffix returns the namespace/name pair given an instance object and a name suffix.
 func NamespacedNameWithSuffix(instance *clv1alpha2.Instance, suffix string) types.NamespacedName {
 	return types.NamespacedName{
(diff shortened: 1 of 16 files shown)


operators/pkg/forge/utils.go
```diff
@@ -77,19 +77,22 @@ func NamespacedNameToObjectMeta(namespacedName types.NamespacedName) metav1.Obje
 	}
 }
 
-// NamespacedNameFromSharedVolume returns the namespace/name pair of the passed SharedVolume.
-func NamespacedNameFromSharedVolume(shvol *clv1alpha2.SharedVolume) types.NamespacedName {
+// NamespacedNameFromObject returns the namespace/name pair of the passed Kubernetes object.
+//
+// Note: Remember all k8s resource structs (e.g. Pod, Deployment, ...) do not implement the
+// metav1.Object interface, but any pointer to them does.
```
AleTopp, line 75: Is this Note okay? @QcFe

AleTopp:
@QcFe Also, I was wondering if you know whether the `forge.NamespacedName` function can be converted to this new utility function too, or the CanonicalName actually does something.
I'm asking since if this function gets an Instance object, I cannot understand why it should have not a canonical name.

Source: https://github.com/netgroup-polito/CrownLabs/blob/7344d9b9fd8bd8411a468eb203aba81eee207719/operators/pkg/forge/utils.go#L59

QcFe:
Ciao @AleTopp, thank you so much for this contribution.
As for your question I admit I'm not sure in which contexts canonicalization is required... I suppose it could be worth trying to see in which places it's used and if it's still needed. In case, I'm not sure if in go now optional parameters are finally supported, but could have been the case to reuse the same function, passing a parameter to enable canonicalization on demand...

AleTopp:
Actually it's used only in one place (besides tests), which is in the examagent so I have no clue about what it does and if the parameter can be uncanonical:
https://github.com/netgroup-polito/CrownLabs/blob/7344d9b9fd8bd8411a468eb203aba81eee207719/operators/pkg/examagent/instance.go#L94

About the second question, no Go doesn't support them, and it never will as it's a design decision.

QcFe:
Ah no in examagent in that context no canonicalization is needed. So feel free to remove it from there (and everywhere else if it's not actually used code).

For sure this kind of canonicalization is used in tenants (like pippo.caio gets a tenant-pippo-caio namespace) but I'd not say it's done by that function directly...

> About the second question, no Go doesn't support them, and it never will as it's a design decision.

👺

AleTopp:
Done!

QcFe (approved):

PR #1135, opened 2026-05-30
Gateway API: Implement HTTPRoute for qlkube and unify domains

SignorMela:
# Description

This PR unifies the frontend and the backend (`qlkube`) under a single domain, significantly simplifying the infrastructure routing by leveraging the Kubernetes Gateway API (`HTTPRoute`). 

Traffic is now routed based on the path (e.g., `/` for the frontend and `/graph` for the backend), leading to a much cleaner and environment-agnostic architecture.

**Main Changes:**
- **Routing:** Moved the `qlkube` mount point in the Express server from the root `/` to `process.env.BASE_URL` (defaulting to `/graph`).
- **Frontend Apollo Client:** Refactored the Apollo Client setup in React. The WebSocket connection URI (`ws://` / `wss://`) is now dynamically parsed and built based on the browser's `window.location` when a relative path is provided.
- **GraphQL Sandbox:** Updated the Apollo Embedded Sandbox script in the Node server to dynamically build the subscription websocket endpoint directly from the browser's origin.
- **Kubernetes / Helm:** 
  - Aligned `qlkube` ingress hostname and TLS secret with the `frontend-app` to share a single TLS certificate.
  - Cleaned up the umbrella chart values (`deploy/crownlabs/values.yaml`) by removing redundancy. 
  - The `qlkube` subchart now safely defaults to the expected `HTTPRoute` logic, consuming the `ingress.path` parameter in a DRY way natively inherited by the deployment via the `BASE_URL` env variable.

### DIFF
--- a/deploy/crownlabs/values.yaml
+++ b/deploy/crownlabs/values.yaml
@@ -17,7 +17,7 @@ frontend-app:
     path: /
   configuration:
     backend:
-      graphql: https://graphql.example.com
+      graphql: /graph
       grafana: https://grafana.example.com/d/example
       bastion: ssh.example.com
     oidc:
@@ -31,6 +31,11 @@ qlkube:
     repository: crownlabs/qlkube
   rbacResourcesName: crownlabs-qlkube
   ingress:
+    hostname: qlkube.crownlabs.example.com
+  httproute:
+    gateway:
+      # name: crownlabs-gw
+      # namespace: envoy-gateway-system
     hostname: qlkube.crownlabs.example.com
   harborToken: {}
   configuration:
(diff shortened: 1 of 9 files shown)


QcFe:
Ciao @SignorMela, thanks for the contribution!
A couple of notes before testing:
- I was thinking it might not be a smart move to remove cors altogether since it'd break local frontend development...
- I'd change the current hostname to omit qlkube and, in the ingress, use the same secret name of the frontend so that the certificate is shared. Plus I'd add a comment in the hostname value saying it's not not used in gwapi mode
- httproute path value could be omitted, sharing the one passed to the deployment

QcFe (approved):

PR #1136, opened 2026-06-03
Refactor PVCProvisioningJob + bugfix

AleTopp:
# Description

This PR refactors the `PVCProvisioningJob` (chown job) by relocating it to a dedicated new package: `storage`. Additionally, it resolves a regression introduced in #947 regarding the PVC `volume-provisioning` (job status) label which was stuck on `pending` even if the job has completed successfully (pointed out by @air-31 on 2026-03-25).

The rationale behind the moving is that the previous location (`utils`) was prone to causing cyclic imports, making the codebase fragile and harder to maintain. This decoupling isolates storage provisioning logic cleanly.

### DIFF
--- a/operators/pkg/controller/sharedvolume/reconciler.go
+++ b/operators/pkg/controller/sharedvolume/reconciler.go
@@ -38,6 +38,7 @@ import (
 	clv1alpha2 "github.com/netgroup-polito/CrownLabs/operators/api/v1alpha2"
 	ctrlcommon "github.com/netgroup-polito/CrownLabs/operators/pkg/controller/common"
 	"github.com/netgroup-polito/CrownLabs/operators/pkg/forge"
+	"github.com/netgroup-polito/CrownLabs/operators/pkg/storage"
 	"github.com/netgroup-polito/CrownLabs/operators/pkg/utils"
 )
 
@@ -232,7 +233,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 		shvolume.Status.PVName = pv.Name
 		shvolume.Status.Phase = clv1alpha2.SharedVolumePhaseProvisioning
 
-		done, err := utils.NFSDriveProvisioning(ctx, log, r.Client, &pvc, &shvolume)
+		done, err := storage.RunPVCProvisioning(ctx, log, r.Client, &pvc, &shvolume)
 		if err != nil {
 			return ctrl.Result{}, err
 		} else if done {
--- a/operators/pkg/forge/storage.go
+++ b/operators/pkg/forge/storage.go
@@ -193,18 +193,13 @@ func MyDrivePVCSpec(storageClassName string, storageSize resource.Quantity) core
 }
 
 // UpdatePVCProvisioningJobLabel updates the provisioning job label for a PVC.
-func UpdatePVCProvisioningJobLabel(pvc *corev1.PersistentVolumeClaim, value string) {
-	if pvc.Labels == nil {
-		pvc.Labels = make(map[string]string)
+func UpdatePVCProvisioningJobLabel(labels map[string]string, value string) map[string]string {
+	if labels == nil {
+		labels = make(map[string]string, 1)
 	}
-	pvc.Labels[ProvisionJobLabel] = value
-}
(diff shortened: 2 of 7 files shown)


AleTopp:
We checked on staging and it works (after re-login) the MyDrive label is updated to `completed`

.golangci.yml
```diff
@@ -81,6 +81,8 @@ linters:
         alias: clctx
       - pkg: github.com/netgroup-polito/CrownLabs/operators/pkg/controller/common
         alias: ctrlcommon
+      - pkg: github.com/netgroup-polito/CrownLabs/operators/pkg/storage-provisioning
+        alias: storage
```
AleTopp: @QcFe do you approve this alias and package name?

QcFe: Think I'd call it storage altogether with no hyphen in the folder too, so no alias is required

operators/pkg/forge/storage.go
```diff
@@ -193,18 +193,13 @@ func MyDrivePVCSpec(storageClassName string, storageSize resource.Quantity) core
 }
 
 // UpdatePVCProvisioningJobLabel updates the provisioning job label for a PVC.
-func UpdatePVCProvisioningJobLabel(pvc *corev1.PersistentVolumeClaim, value string) {
-	if pvc.Labels == nil {
-		pvc.Labels = make(map[string]string)
+func UpdatePVCProvisioningJobLabel(labels map[string]string, value string) map[string]string {
+	if labels == nil {
+		labels = make(map[string]string, 1)
```
QcFe, line 198: What's the behavior if someone later on tries to add another label? If it won't explode then it's fine

operators/pkg/storage-provisioning/provisioning.go
```diff
@@ -26,15 +26,22 @@ import (
...
+// PVCProvisioning enforces a job to provision the passed PVC, changing its owner and adding it a label when done.
+// Returns a bool as first parameter that represents whether the job has completed successfully.
+func PVCProvisioning(ctx context.Context, log logr.Logger, c client.Client, pvc *corev1.PersistentVolumeClaim, owner metav1.Object) (bool, error) {
 	log = log.WithName("provisioning-job")
 
 	val, found := pvc.Labels[forge.ProvisionJobLabel]
 	if !found || val != forge.ProvisionJobValueOk {
-		chownJob := batchv1.Job{ObjectMeta: metav1.ObjectMeta{Name: pvc.Name + "-provision", Namespace: pvc.Namespace}}
+		chownJob := batchv1.Job{
+			ObjectMeta: metav1.ObjectMeta{
```
QcFe: Wasn't there the directive for forging objectMetas with suffix?

operators/pkg/controller/sharedvolume/reconciler.go
```diff
@@ -232,7 +233,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 		shvolume.Status.PVName = pv.Name
 		shvolume.Status.Phase = clv1alpha2.SharedVolumePhaseProvisioning
 
-		done, err := utils.NFSDriveProvisioning(ctx, log, r.Client, &pvc, &shvolume)
+		done, err := storage.PVCProvisioning(ctx, log, r.Client, &pvc, &shvolume)
```
QcFe: Think while we're changing, we can switch to a little more idiomatic name, like, RunPVCProvisioning or something

operators/pkg/storage-provisioning/provisioning.go
AleTopp: Not for provisioning jobs. But will do!

operators/pkg/controller/sharedvolume/reconciler.go
AleTopp: Yea sure, I think we called it this way since you didn't like RunXX at the time😅

QcFe: Eh depends on what the action is. Most of CL has enforce (absence) directives but in case of jobs I suppose it's slightly different

AleTopp:
Tested in staging, everything fine!

QcFe (approved):

PR #1139, opened 2026-06-07
Gateway API: Allowed routes implementation

PaoloGalianoBove:
PR to introduce map for the tenant namespace labels


operators/deploy/operator/values.yaml
```diff
@@ -15,7 +15,10 @@ configurations:
     webhooks: false
     imageList: true
 
-  targetLabel: crownlabs.polito.it/operator-selector=production
+  # Exact-match selector string parsed into labels for tenant namespaces.
+  tenantNamespaceLabels: crownlabs.polito.it/operator-selector=production,crownlabs.polito.it/gw-access=crownlabs-main-production
```
QcFe:
```suggestion
  tenantNamespaceCommonLabels:
    crownlabs.polito.it/instance-resources-replication: 'true'
    crownlabs.polito.it/gw-access: crownlabs-main-production
```

operators/pkg/forge/utils.go
```diff
@@ -139,3 +139,26 @@ func LastCharsOf(s string, many int) string {
 	}
 	return s
 }
+
+// ParseAnnotations parses a string like "key1=val1,key2=val2" into a map.
+func ParseAnnotations(raw string) (map[string]string, error) {
```
QcFe: Think it's better to move this into utils, as a ParseKVMap, or we can keep it here but calling it something like MapFromKVString

operators/pkg/forge/tenant.go
```diff
@@ -45,6 +45,15 @@ func UpdateTenantResourceCommonLabels(labels map[string]string, targetLabel ctrl
 	return labels
 }
 
+// AddStaticTenantNamespaceLabels adds the static labels used to stamp tenant namespaces.
+func AddStaticTenantNamespaceLabels(labels map[string]string) map[string]string {
```
QcFe: Forge names should be semantic, should be read like "forge <something>", so I think we can dropp the "add" in the name :)

operators/pkg/forge/labels.go
```diff
@@ -268,6 +269,16 @@ func TenantLabels(labels map[string]string, tenant *clv1alpha2.Tenant, targetLab
 	return labels
 }
 
+// TenantNamespaceLabels receives in input a set of labels and returns the updated set for a tenant namespace.
+func TenantNamespaceLabels(labels map[string]string, tenant *clv1alpha2.Tenant, tenantCommonNSLabels map[string]string) map[string]string {
+	labels = deepCopyLabels(labels)
+
+	maps.Copy(labels, tenantCommonNSLabels)
+	labels["crownlabs.polito.it/name"] = tenant.Name
```
QcFe: if crownlabs.polito.it/name is used anywhere else I think it could be shared as a global constant

operators/pkg/controller/tenant/suite_test.go
```diff
@@ -118,7 +120,8 @@ var _ = JustBeforeEach(func() {
 		Client:                      cl,
 		Scheme:                      scheme.Scheme,
 		KeycloakActor:               keycloakActor,
-		TargetLabel:                 ctrlcommon.NewLabel("crownlabs.polito.it/operator-selector", "test"),
+		TargetLabel:                 tenantTestTargetLabel,
+		TenantCommonNSLabels:        forge.UpdateTenantResourceCommonLabels(maps.Clone(tenantTestNSLabels), tenantTestTargetLabel),
```
QcFe: Instead of calling forge, I think it'd be better to provide the map directly as an added layer of compliance

operators/deploy/operator/templates/deployment.yaml
```diff
@@ -34,6 +34,7 @@ spec:
           image: "{{ .Values.image.repository }}:{{ include "operator.version" . }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
+            - "--tenant-namespace-common-labels={{ .Values.configurations.targetLabel }},{{ .Values.configurations.tenantNamespaceCommonLabels }}"
```
QcFe: Is this a good approach? What happens if tenantNamespaceCommonLabels is empty? 
Either way I think it'd be better to have this kind of merge at operator logic level (inside the main, right after parsing the common-labels string into a map)...

PaoloGalianoBove: Concatenation has been moved into main.go as requested.

operators/pkg/forge/utils.go
PaoloGalianoBove: Name changed as requested

operators/pkg/forge/tenant.go
PaoloGalianoBove: Name changed as requested

operators/pkg/controller/tenant/suite_test.go
PaoloGalianoBove: Changed as requested

QcFe (approved):

PR #1148, opened 2026-07-01
PMP support for Block volumes mirroring

AleTopp:
# Description

This PR adds support for mirroring of PVs that have `VolumeMode = Block`.
It could have been useful to #1094, but in the end it wasn't actually used.

# How was this tested?

Edited the existing tests to check the added logic.


QcFe:
Ok thanks! I was wondering if you could also add the accessMode check so that it's enforced that the source pvc has RWX...

operators/pkg/controller/pmp/provisioner.go
AleTopp, line 150: As I said during the call, it was already done! @QcFe

QcFe: ahg guh didn't see it

QcFe (approved):

PR #1152, opened 2026-07-06
fix: enable local development and fix frontend/operator schema drift bugs

GiovanniRusso2002:
## Summary

Enables running the full CrownLabs stack (Keycloak, qlkube, frontend, operators) locally, and fixes several real backend/frontend schema mismatches uncovered while setting it up. Full step-by-step local dev guide added.

## What's fixed

**Frontend / GraphQL schema alignment** (bugs affecting any environment, not just local dev):
- Removed `status.ip` from instance queries/mutations/subscriptions — the field no longer exists on `InstanceStatus` (moved to per-environment `status.environments[].ip`), causing `GRAPHQL_VALIDATION_FAILED` on every instance list/create/update.
- Added missing `tenantCrownlabsPolitoItTenantRef` to the instances subscription to match the owning query, fixing an Apollo cache "Missing field" warning on every instance update.
- Fixed `createWorkspace` mutation's `$instances` scalar type to match the currently deployed backend schema (`Int!`).
- Cleaned up a dead `status?.ip` fallback in `utilsLogic.tsx` and hardened template disk-size defaults (`''` → `'0Gi'`) to prevent a crash (`Invalid size string`) when a persistent environment has no disk quota set.

**Operators:**
- `operators/pkg/forge/services.go`: environments of type `Container`/`Standalone` with `guiEnabled: false` produced a Service with zero ports, which Kubernetes rejects (`spec.ports: Required value`). Added a metrics-port fallback, plus test coverage for the previously-untested case.
- `operators/Makefile`: fixed a flag-name typo (`--enable-image-list` → `--enable-imagelist`, which crashed `run-operator` immediately) and disabled `--enable-pmp` by default for local runs (requires a `pvc-mirror` StorageClass not present in lightweight local clusters).
- `operators/samples/templates.yaml` / `workspace.yml`: fixed sample data that no longer passes current CRD validation (environment name length limit, removed `mode` field, missing `disk` size on a persistent environment, `cpu` quoted as string per `resource.Quantity` requirements).

**qlkube (local dev config only, not used in production/Helm deploys):**
- Added missing `tenantV1alpha2Wrapper` / `workspaceWrapperTenantV1alpha2` GraphQL wrappers and a `tenants` subscription entry, without which login/dashboard fail immediately.

**Docs:**
- Added `documentation/local-development.md`: ordered, copy-pasteable guide to run the entire stack locally, including Keycloak realm/client/user setup, port-forwarding pitfalls (pin to a single pod to avoid session-cache issues across HA replicas), operator startup with local storage-class overrides, and a troubleshooting table for every error hit while building this out.

## Test plan
- [x] Local login flow (Keycloak → frontend) works end-to-end
- [x] Instance creation/listing/subscription updates work without GraphQL validation errors
- [x] `go test ./pkg/forge/...` passes (376/376), including new `ServiceSpec` cases
- [x] `tsc -b --noEmit` passes after `graphql-codegen` regeneration
- [x] Verified the GraphQL query fixes also validate against the preprod schema (only the `BigInt`/`Int` operator-version skew differs, expected and handled)


frontend/src/graphql-components/subscription/instances.subscription.ts
```diff
@@ -52,6 +51,17 @@ export default gql`
...
+            name
+            tenantV1alpha2Wrapper {
+              itPolitoCrownlabsV1alpha2Tenant {
+                spec {
+                  firstName
+                  lastName
+                }
+              }
+            }
+          }
```
QcFe, line 56: is this needed?

qlkube/src/index.js
```diff
@@ -24,17 +24,32 @@ dotenv.config();
...
+  function localKubeApiUrl() {
+    return useLocalCluster
+      ? (process.env.LOCAL_K8S_API_URL || 'https://127.0.0.1:6443')
+      : 'http://localhost:8001';
+  }
+  const kubeApiUrl = inCluster ? 'https://kubernetes.default.svc' : localKubeApiUrl();
+  // Talking to the real apiserver (useLocalCluster) requires *some* credential even
+  // for the one-off OpenAPI schema discovery call below, unlike `kubectl proxy` which
+  // used its own kubeconfig regardless. This must be a dedicated, minimally-scoped
+  // token (see documentation/local-development.md), never the end user's own token.
```
QcFe:
```suggestion
  // When targeting a local cluster directly, bypass `kubectl proxy` (which always
  // authenticates with its own kubeconfig credentials and ignores the client's
  // bearer token) and talk straight to the real apiserver, forwarding the token
  // as-is so the apiserver's own OIDC authenticator can enforce per-user RBAC.
  const kubeApiUrl = inCluster ? 'https://kubernetes.default.svc' : 
  useLocalCluster
      ? (process.env.LOCAL_K8S_API_URL || 'https://localhost:6443')
      : 'http://localhost:8001';
```

documentation/local-development.md
QcFe, line 1: If we need to keep this file it's better to translate it and move it into the new dev-local folder

dev-local/keycloak/README.md
```diff
@@ -0,0 +1,276 @@
...
+[`apiserver-oidc-integration.md`](apiserver-oidc-integration.md) for the
+full rationale (why each flag, the `groups`/`aud` claim mappers, etc.) —
+here's just the commands:
+
+```bash
+mkdir -p ~/certs
+kubectl get secret keycloak-tls -n default -o jsonpath='{.data.ca\.crt}' | base64 -d > ~/certs/keycloak-ca.crt
+sudo cp ~/certs/keycloak-ca.crt /etc/rancher/k3s/keycloak-ca.crt
+
+sudo tee /etc/rancher/k3s/config.yaml > /dev/null <<'EOF'
```
QcFe: This is rather dangerous. It's better to tell the user to run `sudo vim /etc/rancher/k3s/config.yaml` and ensure the needed config lines are present.

dev-local/keycloak/README.md
```diff
@@ -0,0 +1,276 @@
...
+unreachable, and any process with an open watch/proxy connection to it
+(`kubectl proxy`, `kubectl port-forward`, the operators) needs restarting
+afterwards. Confirm `kubectl get nodes` reports `Ready` before moving on.
+
+Roll back with:
+
+```bash
+sudo rm /etc/rancher/k3s/config.yaml
+sudo systemctl restart k3s
+```
```
QcFe: not really needed. this is not a tutorial for newbies, people need to understand what they're doing in order to go through this procedure so I'm pretty confident they'd also understand also how to un-do it.

dev-local/keycloak/README.md
```diff
@@ -0,0 +1,276 @@
...
+- ✅ Base `ClusterRole`/`ClusterRoleBinding`s applied.
+- ✅ `mydrive-pvcs` namespace present.
+- ✅ Kubernetes apiserver authenticating via OIDC.
+- ✅ CrownLabs operator running, `Tenant`/`Workspace` reconciled.
+
+For qlkube (talking directly to the real apiserver instead of through
+`kubectl proxy`) and the frontend's HTTPS/OIDC configuration, see
+[`apiserver-oidc-integration.md`](apiserver-oidc-integration.md) Steps 5–7.
+
+## Regenerating the realm export
```
QcFe: This is definitely an interesting section but not strictly needed for the main setup. I'd say it's cleaner and more practical to move into its own dedicated file (keycloak-config-freeze.md or something?)

QcFe:
@GiovanniRusso2002 I rebased from here but it hasn't been a smart move. I'd ask you to:
- rebase locally
- have a look at the pending conversation on the last review
- regenerate the types locally (targeting preprod graphql) since schema could have changed slightly too (this should also fix the failing frontend build)

Also I'd ask @frisso to have a look at the documentation, I'd like your approval too before merging :)

frontend/src/graphql-components/subscription/instances.subscription.ts
GiovanniRusso2002: The tenantV1alpha2Wrapper.itPolitoCrownlabsV1alpha2Tenant.spec.{firstName, lastName} field is needed. It is the only source of tenantDisplayName in utilsLogic.tsx, which is used by RowInstanceTitle.tsx to display the instance owner's name in the manager view (and for sorting and searching).

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
+# Base k3s cluster
+
+The single-node k3s cluster everything else in `dev-local/` (Keycloak,
```
frisso: I cannot understand this sentence. Can you please check?

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## 1. Install k3s
+
+```bash
+curl -sfL https://get.k3s.io | sh -
+```
+
+## 2. kubeconfig access
+
+k3s writes its kubeconfig to `/etc/rancher/k3s/k3s.yaml`, owned by root and
```
frisso: I would kindly ask to add a new line only when the sentence ends with a "dot". Otherwise future reviews will become very difficult.

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
...
+readable only by root by default.
+
+**Don't copy it to `~/.kube/config`** — a copy goes stale the moment k3s
+regenerates its certificates (which it does on its own schedule, not just on
+`systemctl restart`), silently leaving you with an unusable kubeconfig until
+you remember to re-copy it. Instead, tell k3s to write the *original* file
+group-readable, and read it from its real location directly — it's always
+fresh, because it's the actual file k3s keeps updating in place, not a
+snapshot of it.
+
```
frisso, line 19:
```suggestion
This can be done with the following commands:

```

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
...
+```bash
+echo 'export KUBECONFIG=/etc/rancher/k3s/k3s.yaml' >> ~/.bashrc
+export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
+```
+
+### If you already have a kubeconfig with other clusters/contexts
+
+Point `KUBECONFIG` at **both files, live** — not a one-time flattened copy —
+so k3s's own file keeps being read straight from disk, fresh, no matter how
+often its certs get regenerated:
```
frisso:
```suggestion
often its security certificates get regenerated:
```

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```bash
+echo 'export KUBECONFIG=~/.kube/config:/etc/rancher/k3s/k3s.yaml' >> ~/.bashrc
+export KUBECONFIG=~/.kube/config:/etc/rancher/k3s/k3s.yaml
+```
+
+k3s always names its cluster, user and context literally `default`, which
+will very likely collide with an entry already in `~/.kube/config` (e.g.
+Docker Desktop, kind, minikube also default to `default`) — when `KUBECONFIG`
```
frisso:
```suggestion
Docker Desktop, kind, minikube also pick the name `default`).
When `KUBECONFIG`
```

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
...
+Point `KUBECONFIG` at **both files, live** — not a one-time flattened copy —
+so k3s's own file keeps being read straight from disk, fresh, no matter how
+often its certs get regenerated:
+
+```bash
+echo 'export KUBECONFIG=~/.kube/config:/etc/rancher/k3s/k3s.yaml' >> ~/.bashrc
+export KUBECONFIG=~/.kube/config:/etc/rancher/k3s/k3s.yaml
+```
+
+k3s always names its cluster, user and context literally `default`, which
```
frisso: I feel these are very long and convoluted sentences. I would suggest to ask the AI to write in simpler terms: we're creating a tech documentation, not a scientific paper.

dev-local/base-k3s/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```bash
+kubectl config use-context default   # k3s
+```
+
+## Verify
+
+```bash
+kubectl get nodes
+# expected: your node, Ready
```
frisso:
```suggestion
# expected: your node (i.e., the machine k3s is running upon), Ready
```

frisso:
Looks good to me.
Just implement the few modifications in the README, and we're done.

frisso (changes requested):

QcFe (approved):

frisso (approved):

Issue #1143, opened 2026-06-08
[Feature] Migration to Gateway APIs


AleTopp:
This is currently taken on as CloudProg project no.7 (a.y. 2025/26).

## Cloud Programming - Project description
Recently, Kubernetes has deprecated the Ingress API in favor of the newer Gateway APIs. CrownLabs used extensibely Ingress APIs, as each time a new instance (e.g., GUI VM) is created, a new ingress resource is created, which allow the user to connect, using his browser, to the VM. However, given the new API suggested by Kubernetes, and the fact that the current Ingress controller used in CrownLabs has moved to the "unsupported" state, CrownLabs should quickly move its code base to the new APIs.

This project aims at evolving the current CrownLabs codebase to Gateway API, maintaining the compatibility with current applications and services.

### Linked PRs
Cleanup tasks:
- #1104 
- #1105
- #1124 
- #1125 

Main project:
- #1098 
- #1130 
- #1135 
- #1139 
- #1153 
- #1164 
- #1167 
- #1173 
- #1175 
- #1181

PR #1164, opened 2026-07-14
Gateway API: Envoy SecurityPolicy enforcement for Instance authentication

SignorMela:
# Description

This PR migrates Gateway API authentication from legacy NGINX Ingress annotations (`auth-url` and `auth-signin`) to Envoy Gateway's native `SecurityPolicy` resource.

# Key Changes

- **Envoy Gateway support:** Added Envoy Gateway API types and a mock `SecurityPolicy` CRD for `kubebuilder` envtests.
- **Global configuration:** Added the `--gateway-api-auth-service` CLI flag to configure the default authentication endpoint.
- **Controller logic:** The controller parses the `crownlabs.polito.it/auth-service` annotation before creating the `HTTPRoute`. When the annotation is valid, it creates and attaches a `SecurityPolicy`; otherwise, reconciliation stops with an appropriate validation error.

Related to #1143

### DIFF
--- a/deploy/crownlabs/templates/securitypolicy-public-routes.yaml
+++ b/deploy/crownlabs/templates/securitypolicy-public-routes.yaml
@@ -1,0 +1,17 @@
+{{- if and .Values.global.gateway.gatewayApiMode (.Values.global.gateway.securityPolicy.enabled | default true) }}
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: SecurityPolicy
+metadata:
+  name: {{ .Values.global.gateway.securityPolicy.publicRoutes.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "crownlabs.labels" . | nindent 4 }}
+spec:
+  # Intentionally empty to override the global Gateway-level SecurityPolicy and exempt
+  # public routes matching the target selector from authentication.
+  targetSelectors:
+    - group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      matchLabels:
+        {{- toYaml .Values.global.gateway.securityPolicy.publicRoutes.targetSelector | nindent 8 }}
+{{- end }}
--- a/deploy/crownlabs/values.yaml
+++ b/deploy/crownlabs/values.yaml
@@ -8,6 +8,28 @@ global:
     gatewayApiMode: true
     name: "crownlabs-main"
     hostname: crownlabs.example.com
+    securityPolicy:
+      enabled: true
+      name: "crownlabs-global"
+      publicRoutes:
+        name: "crownlabs-public-exclusion"
+        targetSelector:
+          "crownlabs.polito.it/public-route": "true"
+      callback:
+        path: "/app/instauth/callback"
+      secret:
+        enabled: true
+        name: "envoy-secpol-dummy-secret"
+        stringData:
+          client-secret: "dummy"
+      spec:
+        oidc:
+          provider:
+            issuer: "https://auth.ng.crownlabs.polito.it/realms/crownlabs"
+          clientID: "k8s"
+          clientSecret:
+            name: "envoy-secpol-dummy-secret"
+          redirectURL: "%REQ(x-forwarded-proto)%://%REQ(:authority)%/app/instauth/callback"
 
 createClusterRoles: true
 
@@ -306,7 +328,7 @@ gatewayAPI:
   allowedRoutesSimple:
     key: "crownlabs.polito.it/gw-access"
     value: "crownlabs-main-production"
-  # allowedRoutes: {}
+
 
 exam-agent:
   replicaCount: 1
--- a/operators/cmd/instance-operator/main.go
+++ b/operators/cmd/instance-operator/main.go
@@ -182,7 +182,7 @@ func main() {
 	expositionCfg.GatewayAPIMode = gatewayAPIMode
 	log.Info("Gateway API mode selection", "enabled", gatewayAPIMode)
 	if gatewayAPIMode {
-		gwNs, gwName, err := forge.ParseGatewayParent(gatewayAPIRefsValues)
+		gwNs, gwName, err := forge.ParseNamespacedName(gatewayAPIRefsValues)
 		if err != nil {
 			log.Error(err, "invalid gateway parent format, expected 'namespace/name'")
 			os.Exit(1)
--- a/operators/pkg/instctrl/exposition.go
+++ b/operators/pkg/instctrl/exposition.go
@@ -38,23 +38,23 @@ func (r *InstanceReconciler) EnforceInstanceExposition(ctx context.Context) erro
 	instance := clctx.InstanceFrom(ctx)
 
 	if instance.Spec.Running && r.ExpositionConfig.GatewayAPIMode {
-		return r.enforceInstanceExpositionHTTPRoutePresence(ctx)
+		return r.enforceHTTPRoutePresence(ctx)
 	}
 	if instance.Spec.Running && !r.ExpositionConfig.GatewayAPIMode {
-		return r.enforceInstanceExpositionIngressPresence(ctx)
+		return r.enforceIngressPresence(ctx)
 	}
 
 	return r.enforceInstanceExpositionAbsence(ctx)
 }
 
-// enforceInstanceExpositionHTTPRoutePresence ensures the presence of the HTTPRoute required to expose an environment.
-func (r *InstanceReconciler) enforceInstanceExpositionHTTPRoutePresence(ctx context.Context) error {
+// enforceHTTPRoutePresence ensures the presence of the HTTPRoute required to expose an environment.
+func (r *InstanceReconciler) enforceHTTPRoutePresence(ctx context.Context) error {
 	log := ctrl.LoggerFrom(ctx)
 	instance := clctx.InstanceFrom(ctx)
 	environment := clctx.EnvironmentFrom(ctx)
 	envIndex := clctx.EnvironmentIndexFrom(ctx)
 
-	svc, err := r.enforceInstanceExpositionServicePresence(ctx)
+	svc, err := r.enforceServicePresence(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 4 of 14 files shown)


operators/cmd/instance-operator/main.go
```diff
@@ -189,8 +195,33 @@ func main() {
...
+			if err != nil {
+				log.Error(err, "invalid default instances-auth-url format, expected 'namespace/name' when Gateway API mode is enabled")
+				os.Exit(1)
+			}
+			// Create a direct client to fetch the template before the manager starts.
+			cl, err := client.New(mgr.GetConfig(), client.Options{Scheme: rscheme})
+			if err != nil {
+				log.Error(err, "unable to create direct kubernetes client")
+				os.Exit(1)
+			}
```
QcFe:
```suggestion
			cl := client.GetClient()
```

operators/pkg/forge/securitypolicies.go
```diff
@@ -0,0 +1,54 @@
...
+	AuthAnnotationDefault = "default"
+)
+
+// SecurityPolicySpec forges the specification of a Kubernetes SecurityPolicy resource.
+func SecurityPolicySpec(targetRouteName string, templateSpec *egv1alpha1.SecurityPolicySpec) egv1alpha1.SecurityPolicySpec {
+	var spec egv1alpha1.SecurityPolicySpec
+	if templateSpec != nil {
+		spec = *templateSpec.DeepCopy()
+	}
+	spec.PolicyTargetReferences = SecurityPolicyTargetRefs(targetRouteName)
```
QcFe:
```suggestion
	if templateSpec == nil {
		panic("Unexpected empty templatePolSpec")
	}
	spec = *templateSpec.DeepCopy()
	spec.PolicyTargetReferences = SecurityPolicyTargetRefs(targetRouteName)
```

operators/pkg/instctrl/exposition.go
```diff
@@ -222,6 +249,32 @@ func (r *InstanceReconciler) enforceInstanceExpositionServicePresence(ctx contex
...
 }
 
+// enforceInstanceExpositionSecurityPolicyPresence ensures the presence of the SecurityPolicy required to authenticate the HTTPRoute.
+func (r *InstanceReconciler) enforceInstanceExpositionSecurityPolicyPresence(ctx context.Context) error {
+	log := ctrl.LoggerFrom(ctx)
+	instance := clctx.InstanceFrom(ctx)
+	environment := clctx.EnvironmentFrom(ctx)
+
+	targetRouteName := forge.ObjectMetaWithSuffix(instance, environment.Name).Name
+	policy := egv1alpha1.SecurityPolicy{ObjectMeta: forge.ObjectMetaWithSuffix(instance, environment.Name+"-auth-policy")}
```
QcFe:
```suggestion
	policy := egv1alpha1.SecurityPolicy{ObjectMeta: forge.ObjectMetaWithSuffix(instance, environment.Name)}
```
This is a secpol already, no need to tell it in the name too, you're not SignorMelaHuman :)

operators/pkg/instctrl/exposition.go
```diff
@@ -222,6 +249,32 @@ func (r *InstanceReconciler) enforceInstanceExpositionServicePresence(ctx contex
...
+	res, err := ctrl.CreateOrUpdate(ctx, r.Client, &policy, func() error {
+		if policy.CreationTimestamp.IsZero() {
+			policy.Spec = forge.SecurityPolicySpec(targetRouteName, r.ExpositionConfig.GatewayAPISecurityPolicySpec)
+		}
```
QcFe:
```suggestion
	policy.Spec = forge.SecurityPolicySpec(targetRouteName, r.ExpositionConfig.GatewayAPISecurityPolicySpec)
```
the isZero way is something we use mostly only for workloads, to avoid disrupting them in case of changes. However here we need a proper enforcement: if we change the secpol and restart the controller, we need to see all the secpols around updated accordingly.

operators/pkg/instctrl/exposition.go
```diff
@@ -292,6 +349,18 @@ func (r *InstanceReconciler) enforceInstanceExpositionServiceAbsence(ctx context
 	return nil
 }
 
+// enforceInstanceExpositionSecurityPolicyAbsence removes the SecurityPolicy exposed resource for the environment.
+func (r *InstanceReconciler) enforceInstanceExpositionSecurityPolicyAbsence(ctx context.Context) error {
+	instance := clctx.InstanceFrom(ctx)
+	environment := clctx.EnvironmentFrom(ctx)
+
+	policy := egv1alpha1.SecurityPolicy{ObjectMeta: forge.ObjectMetaWithSuffix(instance, environment.Name+"-auth-policy")}
```
QcFe:
```suggestion
	policy := egv1alpha1.SecurityPolicy{ObjectMeta: forge.ObjectMetaWithSuffix(instance, environment.Name)}
```

operators/cmd/instance-operator/main.go
SignorMela: Using mgr.GetClient() before mgr.Start() fails because the cache is not yet started. I simplified it using mgr.GetAPIReader(), which bypasses the cache at boot.

AleTopp: We also had a similar problem, since we wanted to check if the configured StorageClass existed before starting the pmp, but we couldn't Get before the manager started. We solved this by adding a Runnable to the manager, so that it is executed after it starts. Maybe it's applicable also in your case!

https://github.com/netgroup-polito/CrownLabs/blob/f54a14c96ea15145e21c82046b43f848b83bfe4a/operators/cmd/operator/pmp.go#L55

https://github.com/netgroup-polito/CrownLabs/blob/f54a14c96ea15145e21c82046b43f848b83bfe4a/operators/pkg/controller/pmp/provisioner.go#L208-L237

Idk if it's cleaner this or that way @QcFe

QcFe: The APIReader looks fine from my side in this case, it's basically just a cacheless client. I'd avoid overcomoplicating further.
Actually also this pr comment of mine was outdated since we discussed it after I did it. @AleTopp do you see anything bad in using the APIReader?

SignorMela: Thanks @AleTopp! I think APIReader is the cleaner option here. As @QcFe mentioned, it's a simple cacheless client that avoids the boilerplate of a custom Runnable while also allowing us to fail fast at startup if the template is missing.

AleTopp: No I just read the code and thought about our case, nothing against APIReader! :)

QcFe:
As for the policy template, I'd suggest to leave it raw, at least for what concerns the spec. This simplifies the template and enables an even more agnostic approach. Validation is still performed during manifest application so I don't see too much use in pre-enforcing everything...

operators/deploy/instance-operator/templates/clusterrole.yaml
```diff
@@ -45,6 +45,10 @@ rules:
   resources: ["httproutes"]
   verbs: ["get","list","watch","create","patch","update","delete"]
 
+- apiGroups: ["gateway.envoyproxy.io"]
+  resources: ["securitypolicies"]
+  verbs: ["get","list","watch","create","patch","update","delete"]
+
```
QcFe: not needed anymore :)

operators/deploy/instance-operator/templates/_helpers.tpl
```diff
@@ -79,4 +79,27 @@ The tag to be used for image exporter container for VM snapshots
...
+{{- define "instance-operator.instancesAuthRef" -}}
+{{- .Values.configurations.generic.instancesAuthRef | default .Values.configurations.generic.instancesAuthUrl }}
+{{- end }}
+
+{{/*
+The instances authentication URL to be used by the operator
+*/}}
+{{- define "instance-operator.instancesAuthUrl" -}}
+{{- include "instance-operator.instancesAuthRef" . }}
+{{- end }}
```
QcFe: not needed it guess?

operators/deploy/instance-operator/templates/deployment.yaml
```diff
@@ -38,8 +38,9 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - '--namespace-whitelist={{ .Values.configurations.generic.whitelistLabels }}'
-            - '--website-base-url={{ .Values.configurations.generic.websiteBaseUrl | default .Values.global.gateway.hostname }}'
-            - '--instances-auth-url={{ .Values.configurations.generic.instancesAuthUrl }}'
+            - '--enable-auth={{ .Values.configurations.generic.enableAuthentication }}'
```
QcFe: not needed anymore

operators/deploy/instance-operator/values.yaml
```diff
@@ -7,9 +7,11 @@ replicaCountAutomation: 1
 
 configurations:
   generic:
+    enableAuthentication: true
```
QcFe: I'd move this one at deployment level and use it in AND with GWApi mode to decide whether to apply all the secpols

QcFe (commented):
there's some more cleanup to do, also self-review the PR again once you're done to look for possible more leftovers. thanks!

QcFe (approved):

deploy/crownlabs/values.yaml
```diff
@@ -8,6 +8,28 @@ global:
     gatewayApiMode: true
     name: "crownlabs-main"
     hostname: crownlabs.example.com
+    securityPolicy:
+      enabled: true
+      name: "crownlabs-global"
+      publicRoutes:
+        name: "crownlabs-public-exclusion"
+        labelKey: "crownlabs.polito.it/public-route"
+        labelValue: "true"
```
QcFe:
```suggestion
        targetSelector:
          "crownlabs.polito.it/public-route": "true"
```

deploy/crownlabs/templates/securitypolicy-public-routes.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    {{- include "crownlabs.labels" . | nindent 4 }}
+spec:
+  # Intentionally empty to override the global Gateway-level SecurityPolicy and exempt
+  # public routes matching the target selector from authentication.
+  targetSelectors:
+    - group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      matchLabels:
+        {{ .Values.global.gateway.securityPolicy.publicRoutes.labelKey }}: {{ .Values.global.gateway.securityPolicy.publicRoutes.labelValue | quote }}
```
QcFe:
```suggestion
        {{ .Values.global.gateway.securityPolicy.publicRoutes.targetSelector | nindent qualcosa }}
```

PR #1173, opened 2026-07-20
Gateway API: Static Route refinements for the best Ingress migration

alexannekk:
This PR refines the static routing configuration for the frontend, qlkube, and webssh components to ensure their behavior fully mirrors the previous Ingress setup. The adjustments mainly focus on preserving the effects of the original NGINX annotations and aligning paths, so that the Gateway API configuration reproduces the exact same routing semantics as before. 

By restoring parity with the legacy Ingress rules, this work guarantees a seamless migration with no changes in user‑facing URLs and no unexpected deviations in traffic flow.

Related #1143

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,7 +34,7 @@ jobs:
         run: cd operators && make -s getvars | grep GOLANGCI_LINT_VERSION >> $GITHUB_OUTPUT
 
       - name: Check linting
-        uses: golangci/golangci-lint-action@v8
+        uses: golangci/golangci-lint-action@v9
         with:
           version: ${{ steps.golangci-lint-version.outputs.GOLANGCI_LINT_VERSION }}
           working-directory: operators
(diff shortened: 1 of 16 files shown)


alexannekk:
A small note:
These NGINX annotations (`nginx.ingress.kubernetes.io`) are not covered because Envoy Gateway API doesn't expose equivalent capabilities:

- `proxy-buffer-size`: Envoy’s buffering model is fundamentally different. It operates as a streaming proxy and its Gateway does not provide a native mechanism to buffer response headers the way NGINX does. Envoy’s buffering controls apply only to the request body (see envoyproxy/envoy#5368 and https://gateway.envoyproxy.io/docs/tasks/traffic/request-buffering/https://gateway.envoyproxy.io/docs/tasks/traffic/request-buffering/). So, forcing a buffer, is unnecessary and produces a different effect anyway.
- `custom-http-errors`: Gateway API cannot intercept or override upstream HTTP status codes as NGINX does. Envoy can approximate this behavior through EnvoyPatchPolicy, but we suppose it is overwhelming for scenarios like returning a custom 418 response, as in this cases.

QcFe (approved):

QcFe (approved):

PR #1179, opened 2026-07-24
Feat new snapshot controller

abellonioandrea:
# Description

This PR introduces the new architecture for managing Instance Snapshots in CrownLabs, entirely replacing the legacy controller. The new design simplifies the snapshot lifecycle by leveraging CDI direct cloning (PVC-to-PVC).

This PR is related to project 5 about Cloud Programming course ay 2025-26.
Also related to #1172


QcFe:
Ciao @abellonioandrea, thanks for your PR!
Had a brief look at the changes, esp the types. Noticed you switched (esp, added) several things but iirc we eventually converged in keeping this as simple as possible. In particular, the only change I'd suggest is specifying the namespace of the source instance (currently the snapshot had to live in the same instance NS, while iirc we decided to give the user the possiblity to save it elsewhere to target where the new image should live).
Also I remind you that any CRD change needs a manual patch from a cluster admin in order to be tested on the cluster :)

abellonioandrea:
Thanks a lot for your feedback @QcFe
I am working on reducing the complexity and making simpler.
Also thanks for the advice about the CRD changes, I wasn't remembering about it.

PR #1162, opened 2026-07-14
Frontend: managed Disk and GPU (otherResources) quota addition

Rjcky11:
Frontend updates in order to host Disk and GPU quota tracking and modified CPU serialization without changing its type among Template and Workspace files. Added in .env.example a JSON schema which represents custom resources which can be added dinamically to it.

PR #1183, opened 2026-07-31
Documentation Update : resource-quota.md (Follow-up to PRs #1160 and #1162)

Sukuna-shrine:
# Description

Updated doc resource-quota.md with relevant information and changes that were done in PRs #1160 and #1162 regarding disk and gpu quotas, including a step by step guide on how to add new custom resources and an important notice regarding the current frontend UI (included in step 3 of the guide).

Fixes # (issue)

None. It's only documentation.

# How Has This Been Tested?

No testing. It's only documentation.

### DIFF
--- a/operators/docs/resource-quota.md
+++ b/operators/docs/resource-quota.md
@@ -1,71 +1,167 @@
-# Defining quotas per workspace and per user
+# Defining Quotas per Workspace and per User
 
-## Defining quotas per workspace
+This document outlines the architecture, data models, enforcement mechanisms, and instructions for managing resource quotas within CrownLabs.
 
-Each workspace has the maximum resource quota defined in the `spec` of the CrownLabs workspace itself, as in the following example which refers to the `test` workspace:
+---
 
-    admin@k8s-master:~$ kubectl get workspace test -o yaml
-    apiVersion: crownlabs.polito.it/v1alpha1
-    kind: Workspace
-    ...
-    spec:
-      prettyName: CrownLabs workspace dedicated to testing
-      quota:
-        cpu: 10
-        instances: 3
-        memory: 64Gi
-        disk: 15Gi
-        otherResources:
-          nvidia.com/gpu: "1"
-          amd.com/gpu: "1"
+## 1. Defining Quotas per Workspace
 
-In the above example, users belonging to this workspace can launch up to 3 instances (either VMs or containers), but globally cannot consume more than 10CPUs, 64Gi of memory, 15Gi of disk, 1 NVIDIA GPU and 1 AMD GPU.
+Each workspace defines its maximum allowed resource allocation in the `spec.quota` field of the CrownLabs `Workspace` Custom Resource. 
 
-Resource quotas can be changed by simply updating the `quota` section of the workspace specification.
+Example configuration for the `test` workspace:
 
-Types of the resources defined above are the following:
+```yaml
+apiVersion: crownlabs.polito.it/v1alpha1
+kind: Workspace
+metadata:
+  name: test
+spec:
+  prettyName: CrownLabs workspace dedicated to testing
+  quota:
+    cpu: 10
+    instances: 3
+    memory: 64Gi
+    disk: 15Gi
+    otherResources:
+      nvidia.com/gpu: "1"
+      amd.com/gpu: "1"
+```
 
-- CPU: int64
-- Instances: int64
-- Memory: resource.Quantity
-- Disk: resource.Quantity
-- otherResources: map[string]resource.Quantity
+In the example above, users belonging to this workspace can launch up to **3 active instances** (VMs or containers) simultaneously, and globally across the workspace cannot exceed **10 CPUs**, **64Gi of RAM**, **15Gi of Persistent Disk**, **1 NVIDIA GPU**, and **1 AMD GPU**.
 
-## Defining quotas per user
+Resource quotas can be updated at any time by modifying the `quota` section of the `Workspace` resource.
 
-While each CrownLabs workspace has its own resource limits, the values associated to each user depends on the sum of the resource quota of each workspace it belongs to.
-For instance, if a user belongs to a first workspace whose CPU limit is 10, and a second workspace whose CPU limit is 8, the total CPU limit for the user is 18, no matter which workspace it is currently using.
+### Resource Types Specification
 
-In other words, with this (simple) resource quota algorithm, the above user can consume up to 18 CPU even when launching instances (VMs or containers) all within the same workspace, as the user quota has the precedence over the workspace quota.
+The data types used within `ResourceSpec` in the Go backend are defined as follows:
 
-NOTE: Regarding CPU limits it must be taken into account that CPU Cores are essentially summed among different workspaces, but objects like Templates (which belong to a specific Workspace) own an extra property which is ReservedCPUPercentage that limits the CPU Usage of the instances belonging to a specific template with this property defined.
+| Resource Field | Go Data Type | Description |
+| :--- | :--- | :--- |
+| `instances` | `int64` | Maximum number of active/running instances allowed |
+| `cpu` | `int64` | Maximum vCPU cores allocated |
+| `memory` | `resource.Quantity` | Maximum RAM memory (e.g., `16Gi`) |
+| `disk` | `resource.Quantity` | Maximum persistent storage allocated (e.g., `20Gi`) |
+| `otherResources` | `map[string]resource.Quantity` | Generic key-value map for extended hardware accelerators |
 
-Beyond the accumulated limits inherited from course workspaces, each user is allocated a Personal Workspace. This quota acts as a standalone baseline for private workloads and may be configured or expanded by modifying the `personalWorkspace` field directly within the `Tenant`'s Custom Resource.
+---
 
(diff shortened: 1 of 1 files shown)


Sukuna-shrine:
A lint check is failing on a README.md file in the frontend folder (which I didn't modify) due to a Stack Overflow link that fails with a 403 error. How should I proceed? Should I comment out the link? Or remove it? Thanks in advance.

p.s. This time operators failed too because it somehow timed out a test.

operators/docs/resource-quota.md
```diff
@@ -1,71 +1,167 @@
...
+      nvidia.com/gpu: "1"
+      amd.com/gpu: "1"
+```
 
-- CPU: int64
-- Instances: int64
-- Memory: resource.Quantity
-- Disk: resource.Quantity
-- otherResources: map[string]resource.Quantity
+In the example above, users belonging to this workspace can launch up to **3 active instances** (VMs or containers) simultaneously, and globally across the workspace cannot exceed **10 CPUs**, **64Gi of RAM**, **15Gi of Persistent Disk**, **1 NVIDIA GPU**, and **1 AMD GPU**.
```
frisso:
```suggestion
In the example above, users belonging to this workspace can launch up to **3 active instances** (VMs or containers) simultaneously, and cannot exceed **10 CPUs**, **64Gi of RAM**, **15Gi of Persistent Disk**, **1 NVIDIA GPU**, and **1 AMD GPU** globally within the workspace .
```

operators/docs/resource-quota.md
```diff
@@ -1,71 +1,167 @@
...
 
-## Defining quotas per user
+Resource quotas can be updated at any time by modifying the `quota` section of the `Workspace` resource.
```
frisso, line 32:
```suggestion
Resource quotas can be updated at any time by modifying the `quota` section of the `Workspace` resource.
In case resources are decreased, no enforcement is done for all the instances that are already running; hence, users who have already running instances may (temporarily) consume more resources than the maximum allowed values within the workspace.
```

operators/docs/resource-quota.md
```diff
@@ -1,71 +1,167 @@
...
+| Resource Field | Go Data Type | Description |
+| :--- | :--- | :--- |
+| `instances` | `int64` | Maximum number of active/running instances allowed |
+| `cpu` | `int64` | Maximum vCPU cores allocated |
+| `memory` | `resource.Quantity` | Maximum RAM memory (e.g., `16Gi`) |
+| `disk` | `resource.Quantity` | Maximum persistent storage allocated (e.g., `20Gi`) |
+| `otherResources` | `map[string]resource.Quantity` | Generic key-value map for extended hardware accelerators |
 
-Beyond the accumulated limits inherited from course workspaces, each user is allocated a Personal Workspace. This quota acts as a standalone baseline for private workloads and may be configured or expanded by modifying the `personalWorkspace` field directly within the `Tenant`'s Custom Resource.
+---
```
frisso, line 47:
```suggestion
---
`Resource.Quantity` is a common format used in Kubernetes, consisting in a number and a suffix (e.g., 16Gi).
More information [here](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity).
```

operators/docs/resource-quota.md
```diff
@@ -1,71 +1,167 @@
...
+```
 
-1. Environment Phase: The Instance Controller intercepts the quota exhaustion error during the environment enforcement phase and changes the specific environment status to `ResourceQuotaExceeded`.
+---
 
-2. Automatic Recovery (Requeuing): Instead of failing permanently, the controller emits a warning event and automatically schedules a new reconciliation check.
+## 3. How to Add New Custom Resources (Step-by-Step Guide)
 
-Once the tenant terminates other running instances (freeing up resources) or an administrator increases the workspace quota, the next periodic check will automatically succeed, and the environment will transition smoothly into the `Starting` and `Ready` phases.
+CrownLabs uses `otherResources: map[string]resource.Quantity` to avoid hardcoding specific hardware vendor keys in the backend Go codebase. Follow these steps to introduce a new custom resource (e.g., **Intel Gaudi TPU** `intel.com/gaudi` or new **AMD GPUs** `amd.com/gpu`):
```
frisso:
```suggestion
While currently the mostly used _other resource_ are Nvidia GPUs, we cannot exclude that in the future we may need to define new custom resources, such as **Intel Gaudi TPU** `intel.com/gaudi` or new **AMD GPUs** `amd.com/gpu`.

CrownLabs source code has been created in order to support new resources without touching the code itself; in fact, it uses `otherResources: map[string]resource.Quantity` to avoid hardcoding specific hardware vendor keys in the backend Go codebase.

This section documents the steps required to add a new custom resource within CrownLabs, which will be automatically applied in both the frontend (so that workspace administrator can set their limits) and in the backend (by the code that implements resource quota enforcement).
```

operators/docs/resource-quota.md
```diff
@@ -1,71 +1,167 @@
...
+       ▼
+[ 3. ResourceQuota ]      ──(Tenant Namespace Level)───────────► [ Passive Global Security Barrier ]
+```
+
+1. **Validating Webhook (Cluster Level Enforcement):**
+   CrownLabs deploys a `ValidatingWebhookConfiguration` that intercepts `CREATE` and `UPDATE` operations on `Tenants` and `Instances`. 
+   * **Scope Filtering:** Uses `namespaceSelector` to match all CrownLabs-managed namespaces (`workspace-*` and `tenant-*`). This ensures **100% of instance creations are intercepted**, preventing validation bypasses.
+   * **Failure Policy:** Configured with `failurePolicy: Fail` to block unvalidated or over-quota requests before they are written to `etcd`.
+
+2. **ResourceQuota (Tenant Namespace Barrier):**
```
frisso: The ResourceQuota is not clear to me.
1. Which values are set for this resource quota?
2. What's the reason for this mechanism?

operators/docs/resource-quota.md
```diff
@@ -1,71 +1,167 @@
...
+   When a Tenant is reconciled, the Tenant Operator creates a standard Kubernetes `ResourceQuota` inside the user's namespace (`tenant-<username>`). It acts as a passive container limit preventing cluster starvation.
+
+3. **Instance Validator (Workspace Level):**
+   A dedicated Go component that evaluates the real-time consumption of the user against their allowed course workspace limits before an instance is allowed to boot.
```
frisso, line 157: Not clear to me why we need to have these 3 cascading mechanisms.
Can you please clarify?

frisso (changes requested):

frisso:
Dear @Sukuna-shrine ,
thanks for this PR.
I left a couple of comments; I would kindly ask you to clarify, then we're ready for merge.
Thanks!

operators/docs/resource-quota.md
Sukuna-shrine: The three mechanisms operate at different levels of the stack, though the validators work together just at different levels. ResourceQuota normally isn't used if the validators are on, things work like this :

1. Validating Webhook (API Server Gatekeeper): it executes Synchronous pre-checks before objects are written to `etcd`. This provides immediate feedback (e.g., HTTP 422 errors to GraphQL/UI) and evaluates complex CrownLabs business rules (e.g., verifying custom GPU labels, template-level CPU throttles, multi-workspace relationships) which standard Kubernetes primitives cannot understand. To simplify it, it's a check BEFORE the instance is created.

2. Instance Validator (Operator Reconciliation Loop): The difference with the webhook is that it acts at the runtime level. So it's an Asynchronous runtime evaluator inside the Instance Operator controller. It's necessary because it handles race conditions, dynamic runtime updates, and workspace-specific bounds (ex. checking if a specific course workspace quota is respected when launching an instance). It also manages recovery loops (ex. phase shifting to `ResourceQuotaExceeded` and requeuing every 60s during persistent storage allocation failures). To give you an example : if disk creation (PVC), which happens at runtime, fails then the instance validator catches that error and shifts the phase to ResourceQuotaExceeded, after 60 seconds it tries to allocate the disk again, this is a type of error that the webhook validator would not be able to catch since it's something that might happen at runtime.

3. ResourceQuota (fallback) : like i said in the previous comment reply, this is useful if the validator is off it applies some limits to a user if the validator is turned off instead of just leaving it to the native kubernetes limits. The reason why this check only works if the validators are off is because the validator checks are much more restrictive (validator limits resources based on the single workspace you want to create the instance in, ResourceQuota limits resources based on the sum total of all resource limits from all workspaces a user is in (including personal workspace).

if you need any other clarifications or if the document requires some modifications feel free to ask.

operators/docs/resource-quota.md
Sukuna-shrine: 1. The ResourceQuota generated inside a user's namespace is configured with the "aggregated total allowance" of that specific user. This value equals the sum of the quotas across all Workspaces they are enrolled in, plus their Personal Workspace quota (e.g., `hard.cpu = sum(workspace_cpus) + personal_cpu`, `hard.memory = ...`).
2. It serves as a safety net at the namespace boundary. While the custom Webhook and Operators handle CrownLabs-specific business logic at the Custom Resource level (`Tenant`, `Instance`), the native `ResourceQuota` operates directly on low-level Kubernetes primitives (`Pods`, `PVCs`). It ensures that even if custom code experiences an unhandled failure or bypass, the Kubernetes API server natively blocks any pod or container from exceeding the user's hard limits.

frisso (approved):

